Issue #4, opened 2017-12-30
Use better names: currentState, draftState and nextState


mweststrate:

benbraou:
I think this issue has been addressed with commit a3fcf043e26120003852084e4262ae87fd1ab683

mweststrate:
Done

PR #11, opened 2017-12-30
add typescript typings

### LINKED ISSUE #5: Add ts / flow typings

benbraou:
Fixes #5 

Pull request adds typescript definitions only. I used the naming currentState instead of baseState as this seems to be intention of #4


mweststrate:
Thanks a lot!

PR #13, opened 2017-12-31
make isProxy method check if value parameter is defined

### LINKED ISSUE #12: Can't set a defined draft property to undefined
```js
immer(
  {user: 'user'}, 
  draft => {
    draft.user = undefined
  }
)
```
throws `Cannot read property 'Symbol(immer-proxy)' of undefined`

Example --
https://stackblitz.com/edit/react-9qe1eg

benbraou:
Fixes #12 

Add an extra check to isProxy method to handle undefined value

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -170,6 +170,15 @@ describe("base", () => {
         expect(nextState.messy.nested).toBe(baseState.anObject.nested)
     })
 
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks for spotting and fixing! Will release in an iffy

PR #14, opened 2017-12-31
invalidate the proxy of the base state after immer function executes

### LINKED ISSUE #10: Make sure all proxies die after immer func ends
Protects people against accidentally using proxies in async fashion etc

benbraou:
1. Using a revocable proxy instead of a proxy, we invalidate the proxy when immer function executes
1. I used jsdoc @typedef, but I can create an internal typescript definition file and reference it in immer. js if needed

Fixes #10

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -175,10 +175,33 @@ describe("base", () => {
             s.aProp = undefined
         })
         expect(nextState).not.toBe(baseState)
-        expect(baseState.aProp).toBe('hi')
+        expect(baseState.aProp).toBe("hi")
         expect(nextState.aProp).toBe(undefined)
     })
 
+    it("should revoke the proxy of the baseState after immer function is executed", () => {
+        let proxy
+        const nextState = immer(baseState, s => {
+            proxy = s
+            s.aProp = "hello"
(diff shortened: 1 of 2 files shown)


mweststrate:
Hey, Thanks for the PR! Really appreciate it.

Some remarks. I think the PR can be improved in the following way:
1. revoke all the proxies in the `proxies` map. This protects users from passing a (part of) the tree to an async process for example.
2. I would keep the signature of `createProxy` as is, keep the short name and just return the proxy. But, in the `proxies` map store the entire struct to be able to revoke them later on. 

Does that make sense?

I just created a PR for auto-freezing and supporting already frozen trees (that was less trivial then I hoped for). Would you be interested in reviewing it? #15 

Thanks for your contributions!

benbraou:
@mweststrate Thanks for the review! Indeed, that makes sense. I will update my PR accordingly by tomorrow.
Yes, I am interested in reviewing #15 . That will be educational for me. Thanks!

benbraou:
Pull request update as requested. Happy new year!

mweststrate:
Thanks a lot @benbraou !

Issue #2, opened 2017-12-30
Auto freeze in dev mode


mweststrate:

mweststrate:
Released as 0.1.0

PR #15, opened 2017-12-31
Added support for frozen trees, and automatically freeze generated trees

mweststrate:
Implemens #2


immer.js
```diff
@@ -66,7 +79,18 @@ function immer(baseState, thunk) {
         if (isPlainObject(base) || Array.isArray(base)) {
             if (isProxy(base)) return base // avoid double wrapping
             if (proxies.has(base)) return proxies.get(base)
-            const proxy = new Proxy(base, objectTraps)
+            let proxyTarget = base
+            // special case, if the base tree is frozen, we cannot modify it's proxy it in strict mode, clone first.
+            if (Object.isFrozen(proxyTarget)) {
+                proxyTarget = Array.isArray(proxyTarget) ? proxyTarget.slice() : { ...proxyTarget }
```
benbraou: Since spread operator is used, `babel-plugin-transform-object-rest-spread` should be added as a dev dependecny. Also, .babelrc should be updated with `"plugins": ["transform-object-rest-spread"]`

__tests__/frozen.js
```diff
@@ -0,0 +1,49 @@
...
+        array: [1, 2]
+    }
+
+    it("should freeze objects after modifications", () => {
+        expect(Object.isFrozen(baseState.object)).toBe(false) // initially not frozen
+        const next = immer(baseState, draft => {
+            draft.object.c = 2
+        })
+        expect(Object.isFrozen(next.object)).toBe(true)
+        expect(Object.isFrozen(next)).toBe(true)
```
benbraou: Should we also safely assert that `next.array` is not forzen in the process ?

__tests__/frozen.js
```diff
@@ -0,0 +1,49 @@
...
+
+    it("can handle already frozen trees", () => {
+        const a = []
+        const b = { a: a }
+        Object.freeze(a)
+        Object.freeze(b)
+        const n = immer(b, draft => {
+            draft.c = true
+            draft.a.push(3)
+            debugger
```
benbraou: Should debugger statement be removed ?

__tests__/frozen.js
mweststrate: Done

__tests__/frozen.js
mweststrate: Done

immer.js
mweststrate: Good catch, actually I did not put the config there on purpose, to make sure the js files works on modern environments without compilation. I am a bit suprised Jest is fine with this (maybe it is because I switched to node 8). Probably I should add transpilation step soonish...

mweststrate:
@benbraou thanks for the review! Interested in becoming maintainer for this package?

benbraou:
@mweststrate thank you. Yes, I am interested in becoming a maintainer!

PR #16, opened 2018-01-01
fixed typos in readme

Gregjarvez:


immer.js
```diff
@@ -113,10 +113,27 @@ function immer(baseState, thunk) {
...
+
     // create proxy for root
     const rootClone = createProxy(baseState)
     // execute the thunk
-    thunk(rootClone)
+    const maybeVoidReturn = thunk(rootClone)
+
+    //values either than undefined will trigger warning;
+     !Object.is(maybeVoidReturn, undefined) &&
+     console.warn(`Immer callback expects no return value. However %s was returned`, prettify(maybeVoidReturn))
```
mweststrate: Since this is a pretty clear warning, I would keep it simple and just use `typeof maybeVoidReturn` without prettiefying.

yarn.lock
```diff
@@ -32,18 +32,32 @@ acorn-globals@^4.0.0:
   dependencies:
     acorn "^5.0.0"
 
-acorn@^5.0.0, acorn@^5.1.2:
```
mweststrate: any idea where these `yarn.lock` changes are coming from? Maybe a different yarn version (I am on 1.3.2). Since there are no changes in package json, I would expect none here. You can fix it by using`git checkout master yarn.lock`.

mweststrate: Hmm master yarn was not up to date either, just pushed.

immer.js
Gregjarvez: I did that initially . But null ,objects , arrays all printed object . I was thinking I could be more specific. I would use type of instead then

yarn.lock
Gregjarvez: Silly me. I add eslint and prettier to almost every project. I removed it after. That should explain that I will fix it

mweststrate:
Feel free to leave / add prettier setup😊. I have it set up through my ide
atm. I think code is currently formatted with 4 spaces, no semi, double
quote.

Gregjarvez:
Sure. I will do that ;)

Gregjarvez:
I am sorry I will get this done after about an hour. I am not home now

mweststrate:
Merged, thanks a lot!

PR #21, opened 2018-01-02
Remove trailing comma in console.warn

sallyhill:
Avoid unexpected token error from babel compiler


mweststrate:
Will update the prettier config as well, to avoid this in the future

mweststrate:
Released as 0.1.2. Thanks!

PR #22, opened 2018-01-02
Don't require proxies for a correct working

### LINKED ISSUE #8: Create es5 compatible version
If there is enough interest. 

Proxy using defineProperty (the mobx way), use faux arrays and dirty check for key additions

mweststrate:
Fixes #8 

First implementation is promising and feature complete! Needs some clean up, feature detection etc. Performance is quite neat so far (but bare in mind that the Proxy impl is not yet optimized, while the non-proxy impl is)

The update-10k-todos-in-a-set-of-100K benchmark: 

```
    ✓ just mutate (2ms)
    ✓ deepclone, then mutate (475ms)
    ✓ handcrafted reducer (15ms)
    ✓ immutableJS (65ms)
    ✓ immer - with autofreeze (309ms)    // Immer, with auto freeze enabled
    ✓ immer - without autofreeze (148ms) // Immer, but without auto freeze enabled
    ✓ immer - NO PROXIES - with autofreeze (470ms)
    ✓ immer - NO PROXIES - without autofreeze (428ms)
```

### DIFF
--- a/immer-es5.js
+++ b/immer-es5.js
@@ -4,6 +4,7 @@ import {isObject} from "util"
 
 const PROXY_TARGET = Symbol("immer-proxy") // TODO: create per closure, to avoid sharing proxies between multiple immer version
 const CHANGED_STATE = Symbol("immer-changed-state")
+const PARENT = Symbol("immer-parent")
 
 let autoFreeze = true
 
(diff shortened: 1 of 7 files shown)


immer-es5.js
```diff
@@ -0,0 +1,244 @@
+import {isObject} from "util"
```
benbraou: I don't see where isObject is used. Is this a typo?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "immer",
   "version": "0.1.2",
   "description": "Create your next immutable state by mutating the current one",
-  "main": "immer.js",
+  "main": "immer-es5.js",
```
benbraou: main is now  `immer-es5.js`, so the unit tests are testing only immer-es5.js and not immer.js . Is this intended? or may be I am missing something.

benbraou:
Great! 👏
I did not look thoroughly at the implementation details tbh. (I will do it this weekend, but this is independent from this PR of course. For learning purposes 😁)

immer-es5.js
mweststrate: yeah. auto import :(

mweststrate:
@benbraou sorry, forgot to mark it as WIP! Still making some improvements besides the build :)

Issue #7, opened 2017-12-30
Add Travis, batches, coverage, cdn, umd build and such


mweststrate:
- [x] Travis added with f895e016dd4c3374fe5e7f95ff4d642dffbb0367  _thanks @Gregjarvez_
- [ ] batches
- [ ] test coverage
- [ ] cdn 
- [ ] umd build

Gregjarvez:
I've been researching what batches are but, It ain't kicking it. if you wanna give me a run down I can do that tonight ?. And coverage. test coverage? @mweststrate @benbraou

benbraou:
@Gregjarvez yes it is about test coverage
what is done in https://github.com/mobxjs/mobx may help
For the rest, I let @mweststrate clarify

mweststrate:
Yep, build, code style, coverage (run jest with `--coverage` and pipe it through coveralls works pretty great usually), donate badges would be cool. Gitter channel is not needed.

CDN is is simply the unpkg.com link. 

Don't bother about the UMD / ESM builds yet, I first want to figure out what the best way is to export the library in the light of supporting different runtime platforms (with and without proxies, see #22)

mweststrate:
Eh wait, CDN should link to the UMD build, so skip for now :)

Gregjarvez:
Awesome. I will try and work on them tomorrow !. Thanks for the explanation @mweststrate @benbraou

PR #39, opened 2018-01-06
Add travis, prettier and npm badges. Also add unpkg cdn

fkhadra:
Hello,

The changes relate to the issue #7. 

For the CDN part, I only added the unpkg one. For cdnjs the package needs to be registered. I can fill the request to register the package if you want.


mweststrate:
Thanks!

PR #40, opened 2018-01-08
explicit return of setAutoFreeze

Julienng:
Change the definition of `setAutoFreeze` to avoid typescript with `"noImplicitAny": true,` options to throw an Error.


mweststrate:
Thanks!

PR #44, opened 2018-01-08
New immer readme image

Gregjarvez:
it is the same as what you had there initially. just a little bit more presentable
 [Preview image here](https://www.dropbox.com/s/8ad1b7atnfvfdbd/immer_new.png?dl=0)


mweststrate:
Hey @Gregjarvez!

Thanks for taking the effort. However, although neater, I didn't find the new image clearer than the previous one. I also plan to draw some additional images to explain the inner workings of immer, so sticking with the current styles makes that easier :).

Sorry, but thanks for taking the effort though

PR #63, opened 2018-01-13
Improve semiotics in state tree graphic

jkhedani:
Removed a red dot (per original graphic) to illustrate that not all proxies require modification. Third time is the charm! :)


mweststrate:
Thanks!

PR #64, opened 2018-01-13
Adding rollup to build process

### LINKED ISSUE #46: support NODE_ENV='production'  substitution to turn of auto-freeze

arthurdenner:
Hi, @mweststrate.

So, I took a deeper look at microbundle and it seems to not be generating sourcemaps, so I decided to take a chance at building a rollup config myself. Here is my attempt. :sweat_smile: 

If you don't want sourcemaps, I think we can stick with microbundle.

However, to fix #46, I believe we'll need [this plugin](https://github.com/rollup/rollup-plugin-replace) and I don't if it's possible and/or how to add it to microbundle configuration.

PS: I 'm pretty sure this config can be improved, I just don't know how at the moment. If you know how, please let me know. First time using rollup here. :sweat_smile:


mweststrate:
Thanks! will review in a few days! First want to fix the code reuse between the two immer implementations

mweststrate:
I have another PR with probably some conflicting changes, so gonna merge this one, and process the final changes there. Thanks for the PR!

PR #68, opened 2018-01-16
Reuse code between es5 and proxy build

mweststrate:


mweststrate:
Since travis is down; bypassed

PR #70, opened 2018-01-17
chore(build): yarn-or-npm and typings

zewa666:
* do not enforce the use of yarn for build scripts, both yarn and npm should work
* copy typings to dist folder
* reference typings from package.json


mweststrate:
Thanks!

package.json
```diff
@@ -10,11 +10,11 @@
   "types": "./immer.d.ts",
   "scripts": {
     "test": "jest",
-    "test:perf": "yarn build && node --expose-gc node_modules/jest-cli/bin/jest.js --verbose --testRegex '__performance_tests__/.*?js$'",
+    "test:perf": "yarn-or-npm build && node --expose-gc node_modules/jest-cli/bin/jest.js --verbose --testRegex '__performance_tests__/.*?js$'",
```
zewa666: I think this should be called `yarn-or-npm run build`

Issue #60, opened 2018-01-13
minified version


pandanoir:
I want to use minified file, but now immer doesn't have a minified version.

mweststrate:
Umd?

Op za 13 jan. 2018 09:11 schreef Naoto Ikuno <notifications@github.com>:

pandanoir:
yes.
https://unpkg.com/immer/dist/immer.umd.js
This CDN file isn't minified. Many libraries, such as jQuery, lodash and riot.js, provide minified files.

MartinKolarik:
You can use [jsDelivr CDN](https://www.jsdelivr.com/), it supports on-demand minification of any file: https://cdn.jsdelivr.net/npm/immer/dist/immer.umd.min.js

mweststrate:
That is freaking small

Op za 13 jan. 2018 om 13:40 schreef Martin Kolárik <notifications@github.com

pandanoir:
I didn't know that jsDelivr CDN automatically minify and provide a JS file. Thank you!

PR #73, opened 2018-01-17
Use minified CDN link

LukasDrgon:
I replaced the unpkg CDN link with a jsDelivr minified link, to provide a smaller file. #60


mweststrate:
The unpkg build is since last week minified as well, so not needed anymore. Thanks for the PR though!

PR #76, opened 2018-01-19
immer autofreeze set to false during production 

Gregjarvez:


Gregjarvez:
I broke something

mweststrate:
Hey @Gregjarvez 

thanks for the PR! It seems the PR is based on a very old master. Would you be able to reapply the changes to the most recent version of immer?

A few comments on the PR: please keep `autoFreeze` a `let` variable (it can in the latest master be found in `src/common.js`. 

Looking at the autofreeze check, I think your solution to check if the source are minified are pretty nice. So I think the check in the end could be:

```
let autoFreeze = (typeof process != "undefined" && process.env && process.env === "production") || verifyMinified.name === 'verifyMinified'
```

I think the typeof checks are nice to add to make it failsafe against build environments that don't substitute process.env, for example when the build system is not webpack.

Gregjarvez:
sure @mweststrate I will apply those changes soon. however whenever I pull i get merge conflicts and so one although I almost always have not made any changes

mweststrate:
it is probably easier to create a fresh branch based on latest master of
immer and cherry-pick your actual changes

Op za 20 jan. 2018 om 20:37 schreef Coveralls <notifications@github.com>:

Gregjarvez:
Will do that for my next pull request? @mweststrate

mweststrate:
Manually picked the relevant changes, as that would be less work then merging. 

Thanks!

PR #78, opened 2018-01-23
Fixed finalizing array

iruca3:
I got error on this test case. (see the diff)
I could not figure out why this is caused.
It seems some copy of array element is a Proxy.
Therefore, I did finalizing copy on `finalizeArray` in any case.

I don't know the same thing is needed on `finalizeObject`.
Would you review my code please?

I'm very glad to meet such a wonderful library :)

**addition**

I found another case below:
```
const base = {
    arr: [{count: 1}, {count: 2}]
}
const result = produce(base, draft => {
    draft.arr = [
        draft.arr[1]
    ]
    draft.arr[0].count = 1
})
expect(result).toEqual({
    arr: [{count: 1}]
})
```
This also causes an error.
Shouldn't I use the draft object on new array or object....?

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -506,6 +506,40 @@ function runBaseTest(name, useProxies, freeze) {
             })
         })
 
+        it("should not throw error", () => {
+            const base = {
+                arr: [
+                    {
+                        id: "100",
+                        arr: [{id: "1", no: 1}, {id: "2", no: 2}]
+                    },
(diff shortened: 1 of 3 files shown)


mweststrate:
Thanks! I merged it, but implemented it a bit differently as there was another underlying issue. But thanks for finding and reproducing the bug!

mweststrate:
Will release soon as 8.5.0

PR #88, opened 2018-02-02
Fixes #86 creating array proxy

### LINKED ISSUE #86: lodash.find breaks proxy?
`immer` does not modify the draft value on using `lodash.find` on `1.0.0`.

The test code is below:
```
        it("processes with lodash.find", () => {
            const base = [{id: 1, a: 1}, {id: 2, a: 1}]
            const result = produce(base, draft => {
                const obj1 = find(draft, {id: 1})
                const obj2 = find(draft, {id: 2})
                obj1.a = 2
                obj2.a = 2
            })
            expect(result[0].a).toEqual(2)
            expect(result[1].a).toEqual(2)
        })
```

result:
```
    Expected value to equal:
      2
    Received:
      1

      130 |                 obj2.a = 2
      131 |             })
    > 132 |             expect(result[0].a).toEqual(2)
      133 |             expect(result[1].a).toEqual(2)
      134 |         })
      135 |
```

for reproduce: https://github.com/iruca3/immer/tree/bug_with_lodash_find

It reproduces on `1.0.0`, but does not reproduce on `0.8.5`.
I tried to fix this, but I couldn't find out why this is caused.

iruca3:
Fixes #86 :)

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -2,6 +2,7 @@
 import produce, {setAutoFreeze, setUseProxies} from "../src/immer"
 import deepFreeze from "deep-freeze"
 import cloneDeep from "lodash.clonedeep"
+import find from "lodash.find"
 
 jest.setTimeout(1000)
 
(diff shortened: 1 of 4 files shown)


mweststrate:
Thanks! Added some additional tests and merged. Released as 1.0.1. Also fixed #89 :)

PR #98, opened 2018-02-15
Fix TypeScript declarations to support currying and immutable state

tilastokeskus:
The TypeScript declarations prevent modying the produced draft if it was produced from an object that has a readonly interface:

```ts
interface IState {
  readonly foo: boolean;
}

const state: IState = {
  foo: false
};

produce(state, draft => {
  draft.foo = true; // Error: Cannot assign to 'foo' because it is a constant or a read-only property.
});
```

Also calling `produce` with only one argument is not allowed, so currying is impossible:

```ts
const mapper = produce((draft, index) => { // Error: Invalid number of arguments, expected 2.
  ...
});
```

This PR solves both of these cases.

### DIFF
--- a/src/immer.d.ts
+++ b/src/immer.d.ts
@@ -1,3 +1,7 @@
+type Mutable<T extends {[x: string]: any}, K extends string> = {
+	[P in K]: T[P];
+}
+
 /**
  * Immer takes a state, and runs a function against it.
  * That function can freely mutate the state, as it will create copies-on-write.
@@ -12,7 +16,7 @@
  */
 export default function<S = any>(
     currentState: S,
-    recipe: (this: S, draftState: S) => void
+    recipe?: (this: S, draftState: Mutable<S, keyof S>) => void
 ): S
 // curried invocations
 export default function<S = any, A = any, B = any, C = any>(


mweststrate:
thanks for the PR @tilastokeskus! Not sure if this PR really solves the issue. If I understand the typings correctly, I think this will just make the top level type mutable, but not things that are marked as readonly in any subtypes?

tilastokeskus:
@mweststrate You're right, it didn't work properly after all. And apparently there are some other problems too with what I did: https://github.com/Microsoft/TypeScript/issues/13723

From what I gather, recursively removing the readonly-ness from a type properly is not possible at the moment. I pushed a hacky change that makes the state mutable, and at least on my IDE (WebStorm 2018) still gives me correct suggestions when accessing the draft's data. See if you can find glaring issues with it.

ogwh:
I have a simpler fix than this one that also fixes some bugs in the typings. I'm about to create a pull request.

mweststrate:
@tilastokeskus thanks for the effort! I think #109 is a cleaner solution though, so closing this one in favor of that one. N.b. note that you can always use `produce(state, (draft: any) => {})`, so there is no need to actually make that part of the built-in typings.

PR #99, opened 2018-02-17
Improve/fix curried functions typings

kompot:
This PR contains 2 commits.

First one about missing param in a returned function seems like a plain mistyping to me.

Second one might be (?) tricker. Not sure why functionы were sorted from most params to least params.

Given the following code
```
const aReducerPart = produce<IState, IPayload>((draft, action) => {
...
});
```
type of `aReducerPart` was `(IState, IPaylad, any, any) => IState` instead of just `(IState, IPaylad) => IState`.

Reordering curried functions fixed it.

Might break elsewhere in other cases I'm not aware of.


mweststrate:
Thanks!

PR #118, opened 2018-03-07
Add failing tests for recursive use

kachkaev:
Context: https://github.com/mweststrate/immer/issues/100

This PR shows how to get the following error:

>  Cannot use a proxy that has been revoked. Did you pass an object from inside an immer function to an async process?

### DIFF
--- a/__tests__/recursive.js
+++ b/__tests__/recursive.js
@@ -1,0 +1,184 @@
+"use strict"
+import produce, {setUseProxies} from "../src/immer"
+import {isArray, isPlainObject, isString, keys, trim} from "lodash"
+
+const nonPlainObject = () => {}
+
+runTests("proxy", true)
+runTests("es5", false)
+
+function runTests(name, useProxies) {
+    describe("recursive - " + name, () => {
+        setUseProxies(useProxies)
+
+        const testCases = [
+            {
+                title: "in a simple non-recursive case",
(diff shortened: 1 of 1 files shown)


__tests__/recursive.js
```diff
@@ -0,0 +1,184 @@
...
+            }
+        ]
+
+        testCases.forEach(
+            ({title, baseState, desiredState, recursiveTransform}) => {
+                it(`should not leave drafts ${title}`, () => {
+                    const next = recursiveTransform(baseState)
+                    if (desiredState) {
+                        // an exception is thrown during deep comparison if any of the drafts is not finalized
+                        expect(next).toEqual(desiredState)
```
kachkaev: It might also good to add ` expect(next).not.toBe(baseState)` here, but this does not change the idea and the result of the test.

kachkaev:
@mweststrate how easy would it be to fix that failing test? WDYT? 🤔

mweststrate:
I think the example is conceptually broken but didn't have any time yet to
write down why

Op wo 14 mrt. 2018 18:07 schreef Alexander Kachkaev <
notifications@github.com>:

kachkaev:
Sure, no rush. Curious to know what's conceptually broken in the example though  😅 🤔

mweststrate:
Closed in favor of #126

PR #121, opened 2018-03-11
resolve #111: allow passing initial state to curried producer function

### LINKED ISSUE #111: Pass default value as second parameter for curried producer
In redux all reducers should have default value, since redux tree is bootstrapped with `@@redix/INIT` action.

To workaround that, using curried producer as reducer, we have to write condition to check for undefined value of the draft. And we can't use argument's default value `(draft = {count: 0}, action)`, because in this case we don't know if we should return it or able to mutate proxy.
```javascript
const counter = produce((draft, action) => {  
  if (draft === undefined) {
    return {count: 0};
  }

  switch (action.type) {
    case 'INCREMENT':
      return draft.count += 1;
    case 'DECREMENT':
      return draft.count -= 1;
  }
});
```
But writing such check in all producers seems a little overwhelming. What if we could specify default value as a  second parameter of producer, as we do it in `array.reduce` method?
```javascript
const counter = produce((draft, action) => {  
  switch (action.type) {
    case 'INCREMENT':
      return draft.count += 1;
    case 'DECREMENT':
      return draft.count -= 1;
  }
}, {count: 0});
```

Following up #105

pkerschbaum:
Should implement issue #111.

Notes:
- If a function is given both for `fn` and `initialState`, `produce(fn, initialState)` will throw an error, see https://github.com/pkerschbaum/immer/blob/202d54cb96949b45cdfdfb08f180b93723faef29/src/immer.js#L24.
Because if produce would allow this, each call of the curried function with `undefined` would create a curried function again.
- The typings should be improved because currently, `S` is typed with `any`, and if two arguments are passed to `produce`, TypeScript (and I think also Flow) cannot exactly determine which overloaded version of the function gets called.
If we add typings for the state/currentState/initialState `S`, the typing systems should be able to distinguish the `produce` calls.

### DIFF
--- a/__tests__/curry.js
+++ b/__tests__/curry.js
@@ -10,9 +10,6 @@ function runTests(name, useProxies) {
 
         it("should check arguments", () => {
             expect(() => produce()).toThrow(/produce expects 1 or 2 arguments/)
-            expect(() => produce(() => {}, {})).toThrow(
-                /the second argument to produce should be a function/
-            )
             expect(() => produce(new Buffer(""), () => {})).toThrow(
                 /the first argument to an immer producer should be a primitive, plain object or array/
             )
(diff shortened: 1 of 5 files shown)


Anmo:
Hi, I'm starting to test Immer (nice work @mweststrate 👍).

First, thanks @pkerschbaum for this PR.

But, regarding this issue, could I just give my feedback?
I would prefer to have a different method, instead of changing `produce` (even worse, when the original one already receive 2 parameters).

I would do something like (not necessary this method name 😛): `produceWithInitialState(initialState, producer)`.

What you guys think about?

mweststrate:
Great! Thanks a lot

PR #123, opened 2018-03-16
Add badges for npm and gzip size

styfle:
Everyone likes badges so I added some more.
These are kept in sync with npm.

Also I added a link to the JSDelivr CDN which includes some stats on usage.
See https://cdn.jsdelivr.net/npm/immer/


benbraou:
Thanks!

PR #127, opened 2018-03-22
amend flow definitions to include return types from recipe

bugzpodder:
recipe can return a new object.  Add this definition to flowtype


mweststrate:
@bugzpodder the build fails on some flow errors, could you look into those?

bugzpodder:
The only way I managed to fix it was to upgrade flow to 0.68.0.  Any other version I tried (0.65-0.67.1) all had the same build error.

mweststrate:
Cool, thanks!

PR #128, opened 2018-03-23
Migrate source code to TypeScript [WIP]

urish:
- [x] add tsconfig.json
- [x] rename source files .js → .ts
- [x] install types for 3rd party dependencies
- [x] configure build tools for typescript (e.g. jest, rollup, lint-staged, prettier)
- [x] tests pass
- [x] build passes
- [ ] `noImplicitAny` (typewiz)


urish:
@mweststrate can you have a quick look (especially on the few changes introduced in [d08cf3a](https://github.com/mweststrate/immer/pull/128/commits/d08cf3a6ee48147450d9156c97f6012990cf7490)) and let me know if so far it seems good before I go on and start adding types?

Also, are you okay with removing [__tests__/flow/ts.ts](__tests__/flow/ts.ts)? I don't think we need it anymore as the types will be embedded in the source code and verified by ts and by the tests.

src/immer.ts
```diff
@@ -14,7 +14,7 @@ import {produceEs5} from "./es5"
  * @param {Function} producer - function that receives a proxy of the base state as first argument and which can be freely modified
  * @returns {any} a new state, or the base state if nothing was modified
  */
-export default function produce(baseState, producer) {
+export default function produce(baseState, producer?) {
```
styfle: Should these parameters have types defined?

urish: Eventually yes. I think the reason I added the question mark was to try to make `__tests__/flow/ts.ts` happier

mweststrate: The question mark is correct though, produce can be invoked with one argument, basically there are three forms:

```
produce(baseState, producer) => nextState
produce(producer) => baseState => nextState // curried form
produce(producer, initialState) => baseState => nextState // curried form with default state

**Comment by mweststrate** [comment]
@urish looking good so far

**Comment by mweststrate** [comment]
@urish preferrably leave the `ts.ts`, it is not run, but used to verify true negative type check errors

**Comment by urish** [comment]
@mweststrate cool, I'm on it. I created some [initial jest integration for TypeWiz](https://github.com/urish/typewiz/pull/50), though hitting some strange issues when running the instrumented test code - I will have to dig in further and isolate them. It could be that Proxies pose some challenges for TypeWiz, we will find out soon :)

**Comment by urish** [comment]
Update: seems like the Proxies are indeed giving TypeWiz a hard time - the only way I could get all of the tests to pass with TypeWiz was by removing the code that tries to detect the shapes of arrays and object literals. Otherwise, we get all sort of infinite loops and other fun stuff. 

Here are the stats I gathered so far:
- "Stock" TypeWiz - 158 tests fail (66%), tests run for 110 seconds :x:
- TypeWiz without the [object literal shape detection code](https://github.com/urish/typewiz/pull/39) - 44 tests fail (18%) :x:
- TypeWiz without the array types detection code - 157 tests fail (65%) :x:
- TypeWiz without both array / object type detection code - all tests pass in 13.6 seconds :relaxed:

Another challenge is collecting the type information from the test cases when they finish running. I haven't found any straightforward way to do it through the jest plugin, so right now I'm simply adding the following code at the beginning of each test file:

```typescript
afterAll(() => {
    require("fs").writeFileSync(
        "frozen-types.json",
        JSON.stringify((global as any).$_$twiz.get())
    )
})
```
it gets the job done, but then I will still have to merge the collected type info from the different test suites. 

I am not sure how useful this would be without Array / Object type inference - I will give it another try tomorrow, and then I hope to share another report here.

mweststrate:
Closing for inactivity, and since this lib doesn't require much active development TS is not that mandatory

PR #131, opened 2018-03-26
Update immer.d.ts

dmorosinotto:
I changed the return type of recipe to `void | S` to correctly type check the use-case of producers that return new state as shown in the samples of returning data from producers in the readme docs


mweststrate:
Thanks!

dmorosinotto:
You are welcome! 
Hope to see the output of new develop #128 to directly use TS in source code, so maybe in the future you can automatically produce d.ts and keep it in sync with the library 😉

PR #137, opened 2018-04-06
added seamless-immutable to performance tests

pekeler:
had to reduce `MAX` by a factor of 10 so it could complete

### DIFF
--- a/__performance_tests__/add-data.js
+++ b/__performance_tests__/add-data.js
@@ -1,105 +1,103 @@
 "use strict"
 
+import {measure} from "./measure"
 import produce, {setAutoFreeze, setUseProxies} from "../dist/immer.umd.js"
 import cloneDeep from "lodash.clonedeep"
 import {fromJS} from "immutable"
 import Seamless from "seamless-immutable"
 import deepFreeze from "deep-freeze"
 
-describe("loading large set of data", () => {
-    const dataSet = require("./data.json")
-    const baseState = {
-        data: null
-    }
-    const frozenBazeState = deepFreeze(cloneDeep(baseState))
-    const immutableJsBaseState = fromJS(baseState)
-    const seamlessBaseState = Seamless.from(baseState)
+console.log("# add-data - loading large set of data")
 
-    function measure(name, fn) {
-        global.gc && global.gc()
-        test(name, fn)
-    }
+const dataSet = require("./data.json")
+const baseState = {
+    data: null
+}
+const frozenBazeState = deepFreeze(cloneDeep(baseState))
+const immutableJsBaseState = fromJS(baseState)
+const seamlessBaseState = Seamless.from(baseState)
 
-    {
-        const draft = cloneDeep(baseState)
-        measure("just mutate", () => {
-            draft.data = dataSet
-        })
+measure(
+    "just mutate",
+    () => ({draft: cloneDeep(baseState)}),
+    ({draft}) => {
+        draft.data = dataSet
     }
+)
 
-    {
-        const draft = cloneDeep(baseState)
-        measure("just mutate, freeze", () => {
-            draft.data = dataSet
-            deepFreeze(draft)
-        })
+measure(
+    "just mutate, freeze",
+    () => ({draft: cloneDeep(baseState)}),
+    ({draft}) => {
+        draft.data = dataSet
+        deepFreeze(draft)
     }
+)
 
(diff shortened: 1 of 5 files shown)


mweststrate:
@pekeler can you also use https://docs.google.com/spreadsheets/d/1dqiDek-PhuF0yYk_D7iYtY7MzwPLwCy_bLj8gRlh6Bk/edit?usp=sharing and update the README? Preferrably MAX isn't lowered because it might distort all the numbers? Can you also summarize your findings in this thread?

pekeler:
_Summary: seamless-immutable seems to be the slowest when updating many keys (as in the todo test)._

The results on my machine (iMac with 3.5GHz i5) are different from yours, which is why I thought you might want to re-run it and update the graph with your results. Especially deepclone is taking way longer for me.

Here’s my *add-data* result
```
 PASS  __performance_tests__/add-data.js
  loading large set of data
    ✓ just mutate (1ms)
    ✓ just mutate, freeze (10ms)
    ✓ handcrafted reducer (no freeze)
    ✓ handcrafted reducer (with freeze)
    ✓ immutableJS (68ms)
    ✓ immutableJS + toJS (101ms)
    ✓ seamless-immutable (79ms)
    ✓ seamless-immutable + asMutable (88ms)
    ✓ immer (proxy) - without autofreeze (1ms)
    ✓ immer (proxy) - with autofreeze
    ✓ immer (es5) - without autofreeze
    ✓ immer (es5) - with autofreeze
```

Here’s my *todo* result with reduced `MAX`
```
 PASS  __performance_tests__/todo.js (14.358s)
  performance
    ✓ just mutate (4ms)
    ✓ just mutate, freeze (40ms)
    ✓ deepclone, then mutate (331ms)
    ✓ deepclone, then mutate, then freeze (358ms)
    ✓ handcrafted reducer (no freeze) (2ms)
    ✓ handcrafted reducer (with freeze) (2ms)
    ✓ naive handcrafted reducer (without freeze) (1ms)
    ✓ naive handcrafted reducer (with freeze) (42ms)
    ✓ immutableJS (7ms)
    ✓ immutableJS + toJS (67ms)
    ✓ seamless-immutable (5658ms)
    ✓ seamless-immutable + asMutable (5746ms)
    ✓ immer (proxy) - without autofreeze (7ms)
    ✓ immer (proxy) - with autofreeze (12ms)
    ✓ immer (es5) - without autofreeze (30ms)
    ✓ immer (es5) - with autofreeze (30ms)
```

Running *todo* with the original `MAX` takes longer than 15 minutes (don’t know by how much because I lost my patience).

And here’s my *todo* result with original `MAX` and commented-out seamless-immutable tests
```
 PASS  __performance_tests__/todo.js (38.024s)
  performance
    ✓ just mutate (2ms)
    ✓ just mutate, freeze (475ms)
    ✓ deepclone, then mutate (6263ms)
    ✓ deepclone, then mutate, then freeze (6719ms)
    ✓ handcrafted reducer (no freeze) (11ms)
    ✓ handcrafted reducer (with freeze) (23ms)
    ✓ naive handcrafted reducer (without freeze) (15ms)
    ✓ naive handcrafted reducer (with freeze) (485ms)
    ✓ immutableJS (33ms)
    ✓ immutableJS + toJS (611ms)
    ✓ immer (proxy) - without autofreeze (220ms)
    ✓ immer (proxy) - with autofreeze (411ms)
    ✓ immer (es5) - without autofreeze (271ms)
    ✓ immer (es5) - with autofreeze (289ms)
```

pekeler:
Running `test:perf` on a 3.1GHz i7 with node v9.11.1 I see a ton of variance in the measured times. For example, todo's *immutableJS + toJS* test takes anywhere from 800 to 1600 ms, and *immer (es5) - without autofreeze* frequently takes longer(!) than *with autofreeze*.

For this to be a useful benchmark, I think every test needs to be run a few times and then calculate the median value.

mweststrate:
@pekeler thanks for the PR! Is that something you would be willing to add in this PR? Reducing the max, and running the tests a few times? I should be possible to check this repo out and run the tests without needing to skip tests or waiting absurdly long :)

pekeler:
OK, I changed the perf tests to run 9 times each, and then calc and print the median value (sorry - my commit says "mean" which isn't correct).

The perf tests don't use jest any more, which is probably why the coverage decreased.

Jest doesn't support what we need to do.

PR #139, opened 2018-04-08
Replace Object.assign in common, since it's not supported in IE

celebro:
Object.assign() is not available in IE11. In environments without polyfills this library crashes on IE. This PR replaces it with a for-in loop.


scriby:
Can you post benchmarks before and after this change?

mweststrate:
Merged, thanks @celebro!

@scriby tested, doesn't seem have real impact, but might a small change that it doesn't try to polyfill it by default

Issue #138, opened 2018-04-07
Another benchmark


scriby:
I've been doing some experimenting with immer. I really like how much simpler it is to write reducers with it!

I was doing some benchmarking vs ImmutableJS and wanted to check if my results make sense.

The benchmark looks like this:

* Trigger a single action which is handled by one reducer
* The reducer produces a new state by appending the item's id to the end of an array and adding the item into a map (or plain JS object for immer)

I'm triggering the action 3,000 times, and then measuring the performance of triggering the action another 1,000 times.

The immutable JS reducer takes about 400ms, and the immer reducer takes about 4 seconds. Here's the immer profile:

![image](https://user-images.githubusercontent.com/815188/38449255-2b6b0be2-39c1-11e8-9d51-7c8ed0928b87.png)

It looks like the main cost is incurred by producing copies of the state as the state is supposed to be immutable, which has to happen for each triggered action. And that's also the main part that ImmutableJS is able to optimize away.

I just wanted to check with you on whether that basically made sense and whether you had any thoughts on it. (When running immer, proxies are used and freezing is off. Browser is Chrome.)

Thanks,

Chris

mweststrate:
Just checking: Did you test with useProxies true and autoFreeze false? Can
you make a PR out of this benchmark?

Op za 7 apr. 2018 02:42 schreef scriby <notifications@github.com>:

scriby:
Yeah freeze was off and proxies on. I’ll send a PR for the benchmark Monday. Thanks!

scriby:
PR is here: #140

wooandoo:
hi,

immer seems promising.

for information, I did another benchmark: immer vs ramda lens vs shallow copy:
* https://jsbench.me/2mjjchh1pm/1
* https://measurethat.net/Benchmarks/Show/3379/0/immer-setautofreezefalse-vs-shallow-vs-ramda-lens
> same benchmarks, the first one has to be started several times to work, but its results are closest to those I get on NodeJS

with a existing "database" of 100 items (by default):
* immer is 2 times less performant (28,144 ops/s)
* ramda lens ~ shallow copy (40,498 ops/s )
* shallow copy gets the best results (43,369 ops/s)

with a existing "database" of 1000 items:
* immer is relatively better (3,625 ops/s)
* shallow copy is second (4,791 ops/s)
* ramda lens is the better (5,587 ops/s)

with a existing "database" of 10000 items:
* immer (346 ops/s)
[...]

PR #140, opened 2018-04-09
Add incremental benchmark

scriby:
For #138

### DIFF
--- a/__performance_tests__/incremental.js
+++ b/__performance_tests__/incremental.js
@@ -1,0 +1,84 @@
+"use strict"
+import produce, {setAutoFreeze, setUseProxies} from "../dist/immer.umd.js"
+import cloneDeep from "lodash.clonedeep"
+import * as Immutable from "immutable"
+import deepFreeze from "deep-freeze"
+
+function freeze(x) {
+    Object.freeze(x)
+    return x
+}
+
+function createTestObject() {
+    return {
+        a: 1,
+        b: "Some data here"
+    }
+}
+
+describe("Incremental performance", () => {
+    const MAX = 4000
+    const baseState = {
+        ids: [],
+        map: Object.create(null)
+    }
+
+    let immutableJsBaseState
+
+    immutableJsBaseState = {
+        ids: Immutable.List(),
+        map: Immutable.Map()
+    }
+
+    function measure(name, fn) {
+        global.gc && global.gc()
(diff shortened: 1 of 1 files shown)


scriby:
Here are the benchmark results on my machine:

 PASS  __performance_tests__/incremental.js (5.342s)
  Incremental performance
    ✓ just mutate (3ms)
    ✓ handcrafted reducer (no freeze) (1912ms)
    ✓ immutableJS (39ms)
    ✓ immer (proxy) - without autofreeze (2365ms)

mweststrate:
Sorry for not following up yet, definitely thanks for the contribution. Could you add two tests for the full picture: immer in ES5 mode, and immutableJS including a toJS() at the end (I think that is a fairer comparision as in the end in most application stuff has to be toJS-ed, for example during rendering or network requests)

scriby:
No problem adding the ES5 version, but from what I can see I don't think the ImmutableJS.toJS test represents any realistic use case. Serializing back to JS after every action completely defeats the performance gains (and purpose) of using ImmutableJS to begin with.

In most cases, we would just be grabbing individual items out of the ImmutableJS collection, which has negligible overhead. In the case that we want all the items, the best bet would be to iterate over the collection, which is much cheaper than toJS because it can use the items themselves as-is. Perhaps a more realistic benchmark would be to perform all the state changes and then grab some items or iterate over the whole collection?

Based on my understanding of Immer, because it needs to retain immutability of the state, it is basically forced to copy the modified arrays/maps on each update, which is very similar to hand crafted reducers. I think the solution for Immer is similar to the solution for regular redux, which would be something like ImmutableJS, but usable with Immer.

I've actually started hacking on that concept here: https://github.com/scriby/immerutable. Right now I'm working on a Map based on a trie which is very similar to ImmutableJS's one, and a btree for maintaining a collection in sorted order (which ImmutableJS doesn't have an equivalent of). For the same workload as this benchmark, Immerutable is at about 130ms. I'm a couple weeks from finishing/polishing it up, but I think it's looking promising.

Thanks for taking a look,

Chris

jurosh:
Yea I must agree with @scriby, that in real world apps you don't want to ever do `toJS` or `fromJS` as those are most expensive operations. Generally you just work with Immutable pushing there values and iterating still as Immutable without need to wrap/unwrap it.

PR #157, opened 2018-05-27
Fix #148: default returned value

### LINKED ISSUE #148: Default returned value.
It's a bit strange for me:

```
console.log(produce(null, draft => draft)); //null
console.log(produce(null, draft => {})); //undefined

```

stefanwille:
This PR adds a spec for #148. and fixes it. 

When the `baseState` is a primitive value or `null`, immer takes a shortcut and doesn't proxy. But it must still check if the producer returns `undefined`, and if yes, return the `baseState`.

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -905,6 +905,22 @@ function runBaseTest(name, useProxies, freeze) {
             expect(next).toEqual({dots: base.availableStartingDots})
         })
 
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks, nice catch!

mweststrate:
Released as 1.3.1, thanks!

stefanwille:
😄

PR #158, opened 2018-05-29
Fix: Base test uses "if(...)" instead of it(...)"

stefanwille:
There is a test that doesn't run because of a typo: "if(...)" instead of it(...)". This hides other issues in the test.

This PR fixes the typo and also the actual test.


benbraou:
Thank you!

stefanwille:
👍

PR #161, opened 2018-06-09
Add TypeScript Mapped Draft Type

knpwrs:
These changes enable the following:

```ts
import produce from 'immer';
 
interface State {
  readonly x: number;
}
 
// `x` cannot be modified here
const state: State = {
  x: 0;
};
 
const newState = produce<State>((draft) => {
  // `x` can be modified here
  draft.x++;
});
 
// `newState.x` cannot be modified here
```

This helps TypeScript users ensure immutability of their state interfaces. I had to upgrade TypeScript for the `index.d.ts` file to parse properly. This feature was merged into TypeScript back in February: https://github.com/Microsoft/TypeScript/pull/21919

~~It also appears that the `flow/ts.ts` file doesn't run and never did. I tested this externally and it works, but I'd like to see a more robust test suite in this repo.~~

**EDIT:** I've added some TypeScript tests in Jest.

**EDIT:** Fixes #97

PR #170, opened 2018-07-22
Fix TypeScript example

knpwrs:
Of course after #161 was merged I realized that my README example didn't actually do anything. This just makes sure that the original state is passed as the first parameter. Previously `newState` would have been a function instead of the new state object.


mweststrate:
Nice catch :)

PR #172, opened 2018-08-07
fix nested immer calls revoked error

luruozhou:
when outter draft was changed in inner produce calls will cause revoked error
example:

```
    produce({city: {name: "New York"}}, draft => {
                       produce({}, draft2 => {
                           draft.city.name = "shanghai"
                       })
                       draft.city.name
                })
```

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -804,6 +804,20 @@ function runBaseTest(name, useProxies, freeze, useListener) {
             expect(next).not.toBe(bear)
         })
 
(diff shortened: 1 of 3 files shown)


luruozhou:
To be honest, I'm not the user of immer now, maybe in the near future is, but I'm interesting in it.(By the way,I'm using mobx in my project).
As you say, this seems a bit strange to write this way. But I don't know if anyone would write this kind of code carelessly.
 My original intention is to find the potential bug，ignore it if you think this is not necessary.

PR #176, opened 2018-08-18
handle nested ReadonlyArray<T> better when using in TS project

shiftkey:
I was experimenting with `immer` and found myself stuck on how arrays are represented in a TS project.

The example is basically this:

```ts
type ObjectWithReadonlyArray = {
    items: ReadonlyArray<number>
}

function updateReadonlyArray(oldItem: ObjectWithReadonlyArray, newItems: ReadonlyArray<number>) {
    return produce(oldItem, draftState => {
        draftState.items = newItems
    })
}
```

Unfortunately this errors, because `immer` maps both `Array<T>` and `ReadonlyArray<T>` to `DraftArray<T>` which extends Array<T>, and `ReadonlyArray<T>` only has a subset of the `Array<T>` API.

```
Type 'ReadonlyArray<number>' is not assignable to type 'DraftArray<number>'.
  Property 'push' is missing in type 'ReadonlyArray<number>'.
```

I added in some new TS examples to illustrate this, as well as fix the bug in the type declarations.

I think this doesn't require changes to the internals of `immer`, but let me know if there's anything else I can do here.

### DIFF
--- a/__tests__/flow/ts.ts
+++ b/__tests__/flow/ts.ts
@@ -96,3 +96,23 @@ produce([1], draftState => {
         const c: number = a[0];
     }
 })
+
+type ObjectWithArray = {
+    items: Array<number>
+}
+
+function updateArray(oldItem: ObjectWithArray, newItems: Array<number>) {
+    return produce(oldItem, draftState => {
+        draftState.items = newItems
+    })
+}
+
+type ObjectWithReadonlyArray = {
+    items: ReadonlyArray<number>
+}
+
+function updateReadonlyArray(oldItem: ObjectWithReadonlyArray, newItems: ReadonlyArray<number>) {
+    return produce(oldItem, draftState => {
+        draftState.items = newItems
+    })
+}
--- a/src/immer.d.ts
+++ b/src/immer.d.ts
@@ -4,9 +4,13 @@ export type DraftObject<T> = {
   -readonly [P in keyof T]: Draft<T[P]>;
 };
 export interface DraftArray<T> extends Array<Draft<T>> { }
+
+interface DraftReadonlyArray<T> extends ReadonlyArray<Draft<T>> { }
+type DraftArrayish<T> = DraftArray<T> | DraftReadonlyArray<T>
+
 export type Draft<T> =
   T extends any[] ? DraftArray<T[number]> :
-  T extends ReadonlyArray<any> ? DraftArray<T[number]> :
+  T extends ReadonlyArray<any> ? DraftArrayish<T[number]> :
   T extends object ? DraftObject<T> :
   T;
 


shiftkey:
I can see this change breaks some existing tests that expect to be able to build up an array inside the `recipe` callback:

```
    __tests__/types.ts (46,15): Property 'push' does not exist on type 'DraftReadonlyArray<{ readonly value: string; }>'.
    __tests__/types.ts (62,15): Property 'push' does not exist on type 'DraftReadonlyArray<{ readonly value: string; }>'.
    __tests__/types.ts (79,15): Property 'push' does not exist on type 'DraftReadonlyArray<{ readonly value: string; }>'.
```     

After going back to the original issue, here's the shape I have:

```ts
interface IAPIEmail {
  readonly email: string
} class Account {
  public constructor(
    // this is the problem field that blows up when being used inside immer
    public readonly emails: ReadonlyArray<IAPIEmail>,
  ) {}
}
interface IAppState {
  readonly accounts: ReadonlyArray<Account>
}
```

Using the current TS declaration to try and update state manually:

```ts
    const newState = produce(currentState, draftState => {
      draftState.accounts = []
      for (const account of newState.accounts) {
        draftState.accounts.push(account) // error here
      }
    })
```

This is the new error:

```
Argument of type 'Account' is not assignable to parameter of type 'DraftObject<Account>'.
  Types of property 'emails' are incompatible.
    Type 'ReadonlyArray<IAPIEmail>' is not assignable to type 'DraftArray<IAPIEmail>'.
      Property 'push' is missing in type 'ReadonlyArray<IAPIEmail>'.
```

Happy to close this out and open a fresh issue if there's more to this discussion that I haven't quite wrapped my head around.

j-f1:
I wonder if there’s a way to specify that you can set a `ReadonlyArray`, you can then use the set value as an `Array`.

mweststrate:
Didn't read the entire thread, but please not that immer does not support classes. So I can imagine this has impact on the type inference as well for the draft

j-f1:
`ReadonlyArray` isn’t actually a class. `Array`s are cast to `ReadonlyArray`s at the developer’s discretion to indicate they shouldn’t be modified, but they’re still actually regular arrays behind the scenes.

mweststrate:
Sorry, was talking about `Account`

shiftkey:
@mweststrate thanks for the heads up - that wasn't the intent here, but I'll see if I have ideas to get the signature right to prevent that usage.

__tests__/flow/ts.ts
```diff
@@ -96,3 +96,23 @@ produce([1], draftState => {
...
+    })
+}
+
+type ObjectWithReadonlyArray = {
+    items: ReadonlyArray<number>
+}
+
+function updateReadonlyArray(oldItem: ObjectWithReadonlyArray, newItems: ReadonlyArray<number>) {
+    return produce(oldItem, draftState => {
+        draftState.items = newItems
```
mweststrate: The problem is now inverted; because TS no longer knows for sure `.items` is read only or not, this means that this will be a compile error now:
`draftState.items.pop()`

PR #177, opened 2018-08-23
fix: deleting props from object which is in the array should save pat…

luruozhou:
I've found another question, see the code sample below.

```
   let base = [1,2, {name:"jim"}];
   let changes = []

   let fork = produce(
       base,
       draft => {
           delete draft[2].name;
       },
       (patches, inversePatches) => {
           changes.push(...patches)
       }
   )
   console.log(changes) //[]   deleting props from object should save the patches
```


mweststrate:
Closing this PR as it makes the ES5 implementation unnecessary slower, since it diffs now the keyset of arrays. But will add the delete trap as discussed, nice catch!

Issue #153, opened 2018-05-10
Getting the wrapped object from a proxy


RichieAHB:
Please specify:

(Not filling out _all fields_ of this template will probably result in your issue being closed without further notice. If you are not willing fill it in, why would someone else spent his time fixing it?)

* [x] Feature request: I'm currently using immer to modify a subtree inside a larger tree. I want to be able to find nodes in the subtree using reference equality (rather than the predicate I'm currently using due to the proxy), modify that node in the tree then validate the whole tree. Would you be willing to accept a PR that basically allows you to get the wrapped object from the Proxy (in this case a nested proxy, not the root Proxy)? Otherwise maybe a tiny helper like `eq(predicateObj, proxyObj)` that will do the unwrapping and checking?
  * [x] I am willing to implement this myself

RichieAHB:
This is verbatim code but basically it would be nice if `targetPredicate` could be a function that `===` compares two objects. I'd also like to be able to quickly get a reference to the new part in the tree outside of produce (but I'm considering just putting a `Symbol` on the update node in the tree and using that, rather than the original predicate, to find the node again after update). Hope this makes sense!

```
const createProduce = modifier => validator => (
  model,
  targetPredicate,
  ...args
) => dispatch => {
  const newModel = produce(model, draftModel => {
    const targetInfo = find(draftModel, targetPredicate);
    if (!targetInfo) {
      throw new Error(`Couldn't find target`);
    }
    modifier(targetInfo, ...args);
  });

  const errors = validator(newModel);
  const valid = !errors.length;
  // TODO: try and avoid one of these two `find` calls
  const targetInfo = find(newModel, targetPredicate);
  dispatch &&
    dispatch(valid ? newModel : model, valid && targetInfo, !valid && errors);
  return valid;
};
``

**Comment by RoboPhred** [comment]
I came across the need for this as well today.  I am using immer for my internal state management, but I have sub-reducers provided by an external source.  I would like to act like pure-redux to the outside world, but as-is I cannot pass them a non-mutable state.

I essentially have something like this
```
for (const thing of things) {
  const oldState = state.thingStates[thing.id];
  const reducer = thingDef[thing.type].reducer;

  // Issue is here: I want to pass a non-proxy / frozen state to the reducer, as the usage
  // of immer is an implementation detail that the external reducer should not expect or rely on.
  const newState = reducer(oldState || {});

  state.thingStates[thing.id] = newState;
}
```

[...]

PR #179, opened 2018-08-24
Add an original function

RichieAHB:
This addresses #153 and exposes an `original` function, useful for `===` comparing a proxy with an original object. The docs / tests should explain most of the implementation!

### DIFF
--- a/__tests__/original.js
+++ b/__tests__/original.js
@@ -1,0 +1,22 @@
+"use strict"
+import produce, {original} from "../src/immer"
+
+describe("original", () => {
+    const baseState = {
+        a: [],
+        b: {}
+    }
+
+    it("should return the original from the proxy", () => {
+        const nextState = produce(baseState, draftState => {
+            expect(original(draftState)).toBe(baseState)
+            expect(original(draftState.a)).toBe(baseState.a)
+            expect(original(draftState.b)).toBe(baseState.b)
+        })
+    })
+
+    it("should return the same object from an object that is not proxied", () => {
+        const a = []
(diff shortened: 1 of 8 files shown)


__tests__/original.js
```diff
@@ -0,0 +1,22 @@
...
+import produce, {original} from "../src/immer"
+
+describe("original", () => {
+    const baseState = {
+        a: [],
+        b: {}
+    }
+
+    it("should return the original from the proxy", () => {
+        const nextState = produce(baseState, draftState => {
```
mweststrate: could you update the test to check both with `useProxies(true)` and `useProxies(false)`?

__tests__/original.js
```diff
@@ -0,0 +1,22 @@
...
+            expect(original(draftState)).toBe(baseState)
+            expect(original(draftState.a)).toBe(baseState.a)
+            expect(original(draftState.b)).toBe(baseState.b)
+        })
+    })
+
+    it("should return the same object from an object that is not proxied", () => {
```
mweststrate: could you add a test that for a newline introduced object on the draft it returns `undefined`?

mweststrate: And the same for a value that is a primitive

mweststrate:
Looks great! I added some comments about introducing a few more test cases

__tests__/original.js
mweststrate: I think returning `undefined` is slightly more intuitive / better supports fail-fast scenarios better. If you agree, I'll happily merge!

mweststrate:
Merged! Thanks a lot

PR #200, opened 2018-09-20
Upgrade prettier and enable prettier on TypeScript code

stefanwille:
Later versions of Prettier support TypeScript.

This PR upgrades Prettier to the latest version and enables it on TypeScript code.


mweststrate:
Thanks!

PR #206, opened 2018-09-30
Remove superfluous argument: key

styfle:
It looks like createState() only accepts 2 arguments, not 3.

This removes the third argument: `key`


styfle:
@aleclarson Good call! I fixed it 🔧

PR #207, opened 2018-10-02
Add immerutable incremental benchmark.

scriby:
# incremental - lot of small incremental changes

just mutate: 1ms
handcrafted reducer: 485ms
immutableJS: 7ms
immerutable (sorted map): 84ms
immerutable (sorted map) (reversed): 167ms
immer (proxy): 519ms
immer (proxy) (reversed): 2155ms
immer (es5): 4273ms
immer (proxy) - single produce: 6ms
immer (es5) - single produce: 2ms

### DIFF
--- a/__performance_tests__/incremental.js
+++ b/__performance_tests__/incremental.js
@@ -3,17 +3,19 @@ import {measure} from "./measure"
 import produce, {setAutoFreeze, setUseProxies} from "../dist/immer.umd.js"
 import cloneDeep from "lodash.clonedeep"
 import * as Immutable from "immutable"
+import {SortedMapAdapter} from "immerutable"
 
 console.log("\n# incremental - lot of small incremental changes\n")
 
-function createTestObject() {
+function createTestObject(order) {
     return {
         a: 1,
-        b: "Some data here"
+        b: "Some data here",
+        order
     }
 }
 
-const MAX = 1000
+const MAX = 2000
 const baseState = {
     ids: [],
     map: Object.create(null)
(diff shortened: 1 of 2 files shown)


mweststrate:
Hey @scriby 

the relation ship between immer and immerutable is a bit unclear? For example does the latter support deep updates? I'm not against per se against merging another benchmark, especially if it is an extension of immer, but I want to avoid to add benchmarks for all immutability libraries. (Given that there are at least 80 of them). The current benchmarks are just to give an general idea, not to provide an extensive comparison, which should be imho be left to an external, independent party. (Maybe such a thing even exists, didn't check)

scriby:
You could call immerutable an extension of Immer, but it basically just provides some scalable data structures for use with Immer. When used with immer, the data structures become immutable, but they could also be used without Immer and be mutable. They’re specifically designed to have good performance when used with Immer, so there’s not much point using it standalone.

Maybe if you read a little bit more at https://github.com/scriby/immerutable it’ll be more clear.

You’re welcome to merge or not merge this. I just thought you might be interested that there’s a solution to a use case immer is currently weak at (performance of updates to large maps or lists).

Thanks,

Chris

mweststrate:
Yeah, I read the readme (that is were my questions were coming from :-)).
An example in the readme how to combine immer an immerutable, including
deep updates would be a great help. If that looks solid, it could even be
explicitly mentioned in the readme, not just as benchmark

Op do 11 okt. 2018 om 21:21 schreef scriby <notifications@github.com>:

> You could call immerutable an extension of Immer, but it basically just
> provides some scalable data structures for use with Immer. When used with
> immer, the data structures become immutable, but they could also be used
> without Immer and be mutable. They’re specifically designed to have good
> performance when used with Immer, so there’s not much point using it
> standalone.
>
> Maybe if you read a little bit more at
> https://github.com/scriby/immerutable it’ll be more clear.
>
> You’re welcome to merge or not merge this. I just thought you might be
> interested that there’s a solution to a use case immer is currently weak at
> (performance of updates to large maps or lists).
>
> Thanks,
>
> Chris
>
>
> > On Oct 11, 2018, at 11:45 AM, Michel Weststrate <

mweststrate:
@scriby with deep updates I mean, what if the books in your example have a tags collection, and you want to add a tag to a book. How would that look like.

For now I am closing this PR, so didn't really "get" the idea of immerutable, as at first glance it looked like a confusing mix of immutableJS and immer to me, where the data flow isn't really clear. Since there are 70 other immutability libraries with a longer history that aren't part of the performance benchmarks either, I don't see a really clear reason at this point why immerutable should be in the list, despite being inspired by immer. 

If the library get's more widely accepted I have no problem the reconsider, as the performance benefits look promising, but for now I recommend to finetine the readme more first, as I found the code example hard to follow. For example, I didn't understand the concept / role of adapters, or how they interact with `produce`, which seem to be pretty crucial :)

PR #209, opened 2018-10-03
feat: support array insertion and "-" path in `applyPatches`

aleclarson:
WDYT?

(no tests yet)

### DIFF
--- a/src/patches.js
+++ b/src/patches.js
@@ -121,9 +121,35 @@ export function applyPatches(draft, patches) {
                         patch.path.join("/")
                 )
             switch (patch.op) {
-                case "replace":
                 case "add":
-                    // TODO: add support is not extensive, it does not support insertion or `-` atm!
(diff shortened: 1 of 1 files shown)


mweststrate:
The impact is small, let's add it, even though the general goal is not to support everything of json patch. 

Feel free to combine the currently open and approved MRs into a single one if that makes it easier for you

I can give you collaborator permissions as well on this repo, that means that you can create branches and accept merges (after agreeing on them if they are not something like typo's), that makes the whole process around managing MRs easier (as other collaborators can push to those as well)

aleclarson:
Sure, I'll be a collaborator. 😄 

I don't know when I'll get these tests written. Busy working on a library built on top of immer. Turns out this patch isn't crucial for the library, so I can't focus on it now. I *will* be opening at least one more PR that my library needs, and it may not even need tests!

aleclarson:
Closing this for now.

PR #210, opened 2018-10-03
feat: support "remove" of any array index

aleclarson:
WDYT?

(no tests yet)

### DIFF
--- a/src/patches.js
+++ b/src/patches.js
@@ -128,14 +128,21 @@ export function applyPatches(draft, patches) {
                     break
                 case "remove":
                     if (Array.isArray(base)) {
-                        if (key === base.length - 1) base.length -= 1
-                        else
-                            throw new Error(
-                                `Remove can only remove the last key of an array, index: ${key}, length: ${
-                                    base.length
-                                }`
-                            )
-                    } else delete base[key]
+                        const index = Number(key)
+                        if (!Number.isNaN(index)) {
+                            if (index === 0) {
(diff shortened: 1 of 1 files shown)


src/patches.js
```diff
@@ -128,14 +128,26 @@ export function applyPatches(draft, patches) {
...
+                        const index = Number(key)
+                        if (!Number.isNaN(index)) {
+                            if (index < 0 || index > base.length) {
+                                throw new Error(
+                                    "Invalid array patch: The given index is out of bounds"
+                                )
+                            }
+                            if (index === 0) {
+                                base.shift()
+                                break
```
mweststrate: NIT: `if else` ... might make this code block slighly clearer? Extracting utility code might also be nice

mweststrate:
Eh no it doesn't :-). Will accept PR though if you add some tests for the specific cases (the coverage report is a great way to verify that all of them are covered!)

src/patches.js
aleclarson: Probably, I'm just not a fan of a lonesome `break`.

You want `break`s removed and `else`s added? Or keep the `breaks`?

aleclarson: Oh I suppose you mean to break this out into its own function?

mweststrate:
Yes

Op za 13 okt. 2018 02:48 schreef Alec Larson <notifications@github.com>:

aleclarson:
Closing this for now.

PR #211, opened 2018-10-03
improve applyPatches perf

aleclarson:


mweststrate:
Looking good. Did you get any actual numbers to demonstrate the difference?

aleclarson:
I have not done actual benchmarking. Which environments do you care about?

I'd be surprised if a `for..in` isn't faster than `slice + reduce`, but I suppose knowing the magnitude of improvement is valuable.

PR #212, opened 2018-10-03
feat: export `applyPatch` function

aleclarson:
While `applyPatches` is for immutable objects, `applyPatch` is for mutable objects. This would be most useful to Immer-based libraries, but it also lets you apply a JSON patch inside a `produce` recipe function.


mweststrate:
Thanks for the contribution! Would you mind adding adding unit tests, and update the readme and changelog as well? (A `next` section is fine if there isn't one)

aleclarson:
Closing this for now.

PR #216, opened 2018-10-12
Special case for tuples in Draft<T>

### LINKED ISSUE #215: Immer Draft TypeScript type clobbers Tuples
* [x] Issue: Applying the immer `Draft` type to a Tuple type (or anything with a tuple type in it) clobbers the tuple, turning it into an array.
  * Version: 1.7.2
  * _SIMPLE_ Reproduction:
```
// this is a direct copy of the type definitions from immer.d.ts
export type DraftObject<T> = { -readonly [P in keyof T]: Draft<T[P]> }
export interface DraftArray<T> extends Array<Draft<T>> {}
export type Draft<T> = T extends any[]
    ? DraftArray<T[number]>
    : T extends ReadonlyArray<any>
        ? DraftArray<T[number]>
    : T extends object ? DraftObject<T> : T


type tup = [number, number, number, number]

const T: Draft<tup> = [1, 2, 3, 4]
const U: tup = [1, 2, 3, 4]

const TT: tup = T
const UU: Draft<tup> = U
```
or, try it on typescript sandbox
https://goo.gl/zxs3C8
 * Expected behavior: I can assign a tuple of type `Draft<tup>` to a variable of type `tup`
  * Observed behavior: ```Type 'DraftArray<number>' is not assignable to type '[number, number, number, number]'.
  Property '0' is missing in type 'DraftArray<number>'.```
  * [x] Occurs in the ES5 implementation (use `setUseProxies(false)`)

aleclarson:
Fixes #215

Tuples need to be wrapped with `DraftObject` instead of `DraftArray`. Otherwise, I cannot assign a `Draft<T>` to a function expecting a `T`.

### DIFF
--- a/src/immer.d.ts
+++ b/src/immer.d.ts
@@ -3,10 +3,25 @@
 export type DraftObject<T> = {-readonly [P in keyof T]: Draft<T[P]>}
 export interface DraftArray<T> extends Array<Draft<T>> {}
 export type Draft<T> = T extends any[]
-    ? DraftArray<T[number]>
+    ? IsFinite<T> extends true ? DraftObject<T> : DraftArray<T[number]>
     : T extends ReadonlyArray<any>
         ? DraftArray<T[number]>
         : T extends object ? DraftObject<T> : T
+
+export type IsFinite<Tuple extends any[]> = {
+    empty: true
+    nonEmpty: ((..._: Tuple) => any) extends ((
+        _: infer First,
+        ..._1: infer Rest
+    ) => any)
+        ? IsFinite<Rest>
+        : never
+    infinite: false
+}[Tuple extends []
+    ? "empty"
+    : Tuple extends (infer Element)[]
+        ? Element[] extends Tuple ? "infinite" : "nonEmpty"
+        : never]
 
 export interface Patch {
     op: "replace" | "remove" | "add"


arilotter:
In Typescript 3.1, we can replace all the Draft types with just:
`export type Draft<T> = {-readonly [P in keyof T]: Draft<T[P]>};` since they added [mapped types for tuples and arrays ](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#mapped-types-on-tuples-and-arrays).
I guess it'd be a breaking change, though, since we'd lose the exported DraftArray and DraftObject types.

aleclarson:
The `typescript` version that Immer depends on has been incorrect since the `Draft` type was added, because `-readonly` is not supported in TypeScript v2.7 and under.

It's probably best to _not_ use `typesVersions` and just upgrade the `typescript` dependency to the minimum version that fits Immer's needs.

aleclarson:
The final solution:
- turns `Draft<[1]>` into `[1]`
- fixes using `Draft<T>` with functions, `RegExp` objects, dates, and boxed primitives
- has no need for `typesVersions` 🎉 

We'll simplify `IsFinite` out in Immer v2.

PR #219, opened 2018-10-13
Update produce in .d.ts for TypeScript v3+

### LINKED ISSUE #205: Upgraded to 1.7.2, typescript typedefs changes failing build
* [x] Issue: Upgraded to 1.7.2, typescript typedefs changes failing build
  * Version: 1.7.2
  * _SIMPLE_ Reproduction: 

Attempting to build Typescript code that looks like this doesn't work anymore:

```
import {produce, Draft} from 'immer';

const producer = produce((draft: Draft<State>, action: Action) => {
  draft.xyz = action.xyz;
});
```

This code built previously under 1.2.1, but does not build under 1.7.2 (only change is I changed `State` to `Draft<State>`). Typescript is complaining that "void" is not assignable to type DraftObject<State>.

I think it's supposed to be matching this type def:

```
// 1 additional argument of type A
<S = any, A = any>(recipe: (this: Draft<S>, draftState: Draft<S>, a: A) => void|S): 
    (currentState: S, a: A) => S
```

I think the issue is that the typedef declares that the recipe function will return an "S", but that's going to be difficult because it receives a `Draft<S>`. Tweaking the type def locally and having recipe return `void|Draft<S>` fixes the issue for me.

  * Expected behavior: It builds
  * Observed behavior: It doesn't build
  * [x] Occurs in the ES5 implementation (use `setUseProxies(false)`)

aleclarson:
Fixes #205

I don't think this works under TypeScript 3.0 (or even 3.1)

### DIFF
--- a/src/immer.d.ts
+++ b/src/immer.d.ts
@@ -30,84 +30,29 @@ export interface IProduce {
      * @param initialState - if a curried function is created and this argument was given, it will be used as fallback if the curried function is called with a state of undefined
      * @returns The next state: a new state, or the current state if nothing was modified
      */
-    <S = any>(
+    <S = any, R = never>(
         currentState: S,
-        recipe: (this: Draft<S>, draftState: Draft<S>) => void | S,
+        recipe: (this: Draft<S>, draftState: Draft<S>) => void | R,
         listener?: PatchListener
-    ): S
+    ): R
 
     // curried invocations with default initial state
     // 0 additional arguments
-    <S = any>(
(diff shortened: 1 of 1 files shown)


aleclarson:
I wonder if you can have a `.d.ts` for a specific version range of TypeScript.

arilotter:
@aleclarson you can, check the `typesVersions` section here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html

aleclarson:
Moved to the `v2` branch

If you can get `typesVersions` working with TypeScript v2.x, let me know.

PR #220, opened 2018-10-16
feat: export `isDraft` function

aleclarson:
To use in other libraries


aleclarson:
I guess you *could* do `!!original(value)` to get the same effect as `isDraft(value)`. WDYT?

mweststrate:
Should isDraft return true for a new object that was added to the draft
tree? Probably not?

Op vr 19 okt. 2018 20:51 schreef Alec Larson <notifications@github.com>:

aleclarson:
@mweststrate "Just proxies" is the idea. LGTM?

mweststrate:
@aleclarson LGTM. Could you add some tests for both proxy and non proxied version?

aleclarson:
Are tests necessary? It's a very simple function: `const isDraft = (value) => !!value && !!value[PROXY_STATE]`

mweststrate:
For regression against internal refactorings, yeah please provide a some
false and positive tests to protect against accidental mistakes.

Op zo 21 okt. 2018 12:34 schreef Alec Larson <notifications@github.com>:

aleclarson:
Done!

PR #221, opened 2018-10-16
feat: export `setHook` function

aleclarson:
Adds the following hooks:

- `setProperty(draftState, prop, newValue)`
- `deleteProperty(draftState, prop)`
- `copyObject(draftState)`

```ts
import {setHook} from 'immer'

setHook('setProperty', (state, prop, newValue) => {
  console.log(prop, '=', newValue)
})
```

Currently, `setHook` can only be called **once** per hook.

These hooks are nice for tracking copies and changed properties.


mweststrate:
I'm currently preparing for a conf, so postponing on reviewing, non-trivial MRs, will get back at this in two weeks!

mweststrate:
The goal of this PR is unclear to me :)

aleclarson:
@mweststrate Provide hooks for other packages to use. These are intended for library authors. They could also be used for some kind of Immer debugger, if anyone ever finds a reason for doing so. ;)

aleclarson:
Bummed we couldn't fit this into v1.8.0 :(

aleclarson:
I have two packages in the works where it's crucial (perf-wise) for them to plug into Immer's draft processing phase. If this PR isn't merged, they'll be forced to traverse every `produce`-returned copy manually and they would miss out on the rich metadata provided by draft objects. 💯 🔥

aleclarson:
Closed in favor of #258

@mweststrate When the libraries are ready, I will point you to where I'm using Immer hooks.

aleclarson:
@mweststrate I'm not really sold on the "unsafe_" prefix rationale. If the contract needs amendment, publish a major version. Why not? You could even do a pre-release if you want to gather feedback before finalizing the API. This way, you're not threatening users of the API with breakage for (seemingly) no reason.

PR #222, opened 2018-10-16
fix: ensure all properties exist on es5 object copies

aleclarson:
Previously, getters and non-enumerable keys were never copied


aleclarson:
I'm closing this one. I think I can get by without it.

PR #227, opened 2018-10-22
fix: use proxies only if Reflect exists

### LINKED ISSUE #226: Use es5 implementation in browsers where Proxy is supported but Reflect is not
* [ ] Feature request: ______
  * [x] I am willing to implement this myself
* [x] Issue: Browsers in which Proxy is not supported ES5 implementation is used but in `firefox 41` `Proxy` is supported but `Reflect` is not so it default implementation with Proxies is used which throws error `Reflect is not defined`
  * Version: 1.7.1
  * _SIMPLE_ Reproduction: codesandbox itself doesn't work in `firefox 41` but check of support of [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#Browser_compatibility) and [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect#Browser_compatibility) confirms the issue.
  * Expected behavior: ___
  * Observed behavior: ___
  * [ ] Occurs in the ES5 implementation (use `setUseProxies(false)`)

Fix: Check for support for both Reflect and Proxy to use ES5 implementation.

aleclarson:
Fixes #226


src/common.js
```diff
@@ -20,7 +20,7 @@ const inProduction =
     verifyMinified.name !== "verifyMinified"
 
 let autoFreeze = !inProduction
-let useProxies = typeof Proxy !== "undefined"
+let useProxies = typeof Reflect !== "undefined"
```
mweststrate: please check both Reflect and Proxy :)

aleclarson: To guard against rare environments? According to MDN, `Reflect` is never available where `Proxy` isn't on the major browsers.

vikas-parashar: @mweststrate This would work?
```
// Use Proxy only when both Proxy and Reflect is supported.
let useProxies = typeof Proxy !== undefined && typeof Reflect !== undefined
```

@aleclarson are you planning to fix this or do you want me to create a new PR with this fix? :)

aleclarson:
@mweststrate Just merged this. Should we cut a release?

mweststrate:
@aleclarson yes, thanks! I can give you npm publish rights if you send me your npm username :-)

aleclarson:
@mweststrate https://www.npmjs.com/~aleclarson

PR #231, opened 2018-10-25
fix: never auto-freeze new objects

### LINKED ISSUE #229: Should produce always freeze its return value?
Which guarantee is more important?
- `produce` always freezes objects it returns
- `produce` is a no-op when the producer is a no-op

Have a look at the following behavior: (https://codesandbox.io/s/1o8x91w5l4)

```ts
let obj: object

// 1. do nothing with draft of a mutable object
Object.isFrozen(
  produce({}, () => undefined)
) // => false

// 2. return the draft of a mutable object
Object.isFrozen(
  produce({}, draft => draft)
) // => false

// 3. replace null with a mutable object
obj = {}
Object.isFrozen(
  produce(null, () => obj)
) // => false

// 4. replace a mutable object with a mutable object
obj = {}
Object.isFrozen(
  produce({}, () => obj)
) // => true

// 5. return the original, mutable object
obj = {}
Object.isFrozen(
  produce(obj, () => obj)
) // => true

// 6. your average producer
Object.isFrozen(
  produce({}, draft => {
    draft.a = 1
  })
) // => true
```

Which of these cases have unexpected behavior?

**Case 1** seems wrong if you believe `produce` should always freeze objects before returning them.

**Case 2, 3** seem wrong. Why shouldn't the return value be frozen here?

**Case 4, 5, 6** seem right. The return value is frozen when it's an object.

@mweststrate WDYT?

aleclarson:
Closes #229 

This passes the tests added by a60a6fa08f0015ca66d8e3161a8197982fdba417

@mweststrate Is this considered breaking?

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -867,22 +867,6 @@ function runBaseTest(name, useProxies, freeze, useListener) {
             expect(next.user).toEqual(user)
         })
 
-        if (freeze)
-            it("should freeze new data well", () => {
-                const base = {}
-                const next = produce(
-                    base,
-                    draft => {
-                        draft.x = {y: [{z: true}]}
-                    },
-                    listener
-                )
-                expect(Object.isFrozen(next)).toBe(true)
-                expect(Object.isFrozen(next.x)).toBe(true)
-                expect(Object.isFrozen(next.x.y)).toBe(true)
-                expect(Object.isFrozen(next.x.y[0].z)).toBe(true)
-            })
-
         it("should structurally share identical objects in the tree", () => {
             const base = {bear: {legs: 4}, eagle: {legs: 3}}
             const next = produce(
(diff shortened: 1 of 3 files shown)


mweststrate:
Note, I'm travelling next two weeks, so it might take a while before I have time to review, should really be focusing on prepping talks now :)

aleclarson:
@mweststrate Are you using "minor" literally? Or do you mean 1.7.5? :)

mweststrate:
I think 1.8, don't have strong arguments though :-)

mweststrate:
Ah sorry, missed that! feel free to revert the merge and merge to 1.8
instead :-)

Op vr 9 nov. 2018 om 12:33 schreef Alec Larson <notifications@github.com>:

PR #232, opened 2018-10-26
Improve Flow types

migueloller:
Adds better Flow support. Now, if the return type of the recipe never returns `undefined`, it can be guaranteed that produce won't return `undefined`. This allows compatibility with other types, like Redux's `Reducer` type.

i.e.,

```es6
import type { Reducer } from 'redux'

type State = {| count: number |}

type Action = {| type: 'INCREMENT' |} | {| type: 'DECREMENT' |}

const reducer: Reducer<State, Action> = (state: State = { count: 0 }, action) => {
  switch(action.type) {
    case 'INCREMENT':
      state.count++
      break

    case 'DECREMENT':
      state.count--
  }

  return state
}

// This is now possible! Before it would fail, as the return type of
// produce would be `State | void`.
const initialState: State = produce(reducer)(undefined, { type: 'INCREMENT' })
```

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,4 +5,5 @@
 __tests__/flow
 
 [options]
+include_warnings=true
 suppress_comment= \\(.\\|\n\\)*\\$ExpectError
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "watch": "jest --watch",
     "test": "jest",
     "test:perf": "NODE_ENV=production yarn-or-npm build && cd __performance_tests__ && babel-node add-data.js && babel-node todo.js && babel-node incremental.js",
-    "test:flow": "yarn-or-npm flow check",
+    "test:flow": "yarn-or-npm flow check --max-warnings=0",
     "coveralls": "jest --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
     "build": "rimraf dist/ && cross-env NODE_ENV=production rollup -c && cpx \"src/immer.{d.ts,js.flow}\" dist",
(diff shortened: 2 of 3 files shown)


migueloller:
Seems to be failing tests that fail in `master` as well. Looks like #231 fixes it. Will rebase once it's merged into master.

aleclarson:
You shouldn't need to rebase, since you're not touching any of the same files as #231

migueloller:
@aleclarson, sweet! Let me know if there's anything else you need from 😄

mweststrate:
The CI failed, I guess this still needs rebasing?

aleclarson:
@mweststrate This PR was put up before I removed failing tests from master with a rebase (naughty, I know).

migueloller:
@mweststrate, thanks for the review! Rebasing now.

mweststrate:
@migueloller, some tests are failing in the CI. Probably locally as well. Could you look into this?

migueloller:
@mweststrate, I spent a bit looking at the failing test and realized a few things.

- There are other regressions on the Flow types
- `$ExpectError` tests weren't even working at all
- The currently proposed types seem to break down in some use cases

I want to spend some more time to address these issues but I'm extremely busy with my day job at the moment. I will push new changes ASAP.

mweststrate:
Merged into 3.0 branch

migueloller:
@mweststrate, should we reopen as it [was removed](https://github.com/immerjs/immer/pull/309#issuecomment-483341908) from the 3.0 branch?

mweststrate:
Undid that :)

mweststrate:
Closing for inactivity

PR #236, opened 2018-10-29
fix(produce): avoid throwing on unproxyable objects

### LINKED ISSUE #233: Dates and RegExps are neither primitives nor plain objects
`Date` and `RegExp` objects cannot be passed to `produce` without an error being thrown.

https://codesandbox.io/s/rrz9lq7pwp

aleclarson:
Fixes #233 

I'm thinking this is a breaking change, since someone may be relying on an error being thrown in some weird situation.


aleclarson:
Tests added 👍

PR #238, opened 2018-10-29
fix(produce): never curry when 2nd argument is a function

aleclarson:
Currently, the first 2 arguments of `produce` cannot both be a function. Otherwise, an error is thrown.

Instead of throwing an error, the `produce(baseState, recipe)` code path should be taken, which would let users use functions as a base state. This goes really well with #236, because it lets us guarantee to callers that _any value_ can be used as the base state of a `produce` call.

You _cannot_ use a function as the `initialState` of a `produce(recipe, initialState)` call (also known as: a curried producer).

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -1275,6 +1275,15 @@ function runBaseTest(name, useProxies, freeze, useListener) {
             })
         })
 
+        it("allows a function as the base state", () => {
+            let fn = () => {}
+            expect(
+                produce(fn, draft => {
+                    expect(fn).toBe(draft)
+                })
+            ).toBe(fn)
+        })
+
         it("cannot return and produce undefined!", () => {
             const base = 3
             expect(produce(base, () => 4)).toBe(4)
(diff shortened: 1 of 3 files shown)


mweststrate:
@aleclarson good catch! Would you mind adding a unit test?

aleclarson:
Yeah, I can get the tests figured out. Do you consider this a breaking change, or should we move this to next patch?

mweststrate:
I think this is just a fix and could be done in a patch

aleclarson:
Strange, my local `jest --coverage` reports no decrease in coverage:
```sh
------------|----------|----------|----------|----------|-------------------|
File        |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------|----------|----------|----------|----------|-------------------|
All files   |    97.18 |    94.81 |    97.06 |    97.82 |                   |
 common.js  |    94.12 |    90.48 |    88.89 |    93.22 |       74,75,76,79 |
 es5.js     |    98.46 |     96.7 |      100 |    99.13 |               248 |
 immer.js   |      100 |      100 |      100 |      100 |                   |
 patches.js |    98.04 |    95.24 |      100 |      100 |                89 |
 proxy.js   |    96.47 |    93.94 |      100 |     97.5 |           156,157 |
------------|----------|----------|----------|----------|-------------------|
```

PR #249, opened 2018-11-09
simplify es5 object key diffing

aleclarson:
The `shallowEquals` function you took from `fbjs` had a branch that wasn't used by Immer, so I merged it into `hasObjectChanges` and even optimized it. :)


mweststrate:
Cool! Thanks :)

PR #256, opened 2018-11-27
perf: proxy assign

aleclarson:
Optimize based on value's truthiness. Truthy values are guaranteed to never be undefined, so we can avoid the `in` operator. Lastly, truthy values may be proxies, but falsy values are never proxies.

### DIFF
--- a/src/proxy.js
+++ b/src/proxy.js
@@ -90,16 +90,17 @@ function get(state, prop) {
 }
 
 function set(state, prop, value) {
-    // TODO: optimize
-    state.assigned[prop] = true
     if (!state.modified) {
-        if (
-            (prop in state.base && is(state.base[prop], value)) ||
-            (has(state.proxies, prop) && state.proxies[prop] === value)
-        )
-            return true
+        // Optimize based on value's truthiness. Truthy values are guaranteed to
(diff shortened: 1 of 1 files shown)


src/proxy.js
```diff
@@ -90,16 +90,17 @@ function get(state, prop) {
...
+        // Optimize based on value's truthiness. Truthy values are guaranteed to
+        // never be undefined, so we can avoid the `in` operator. Lastly, truthy
+        // values may be proxies, but falsy values are never proxies.
+        const isUnchanged = value
+            ? is(state.base[prop], value) || value === state.proxies[prop]
+            : is(state.base[prop], value) && prop in state.base
+        if (isUnchanged) return true
         markChanged(state)
     }
+    state.assigned[prop] = true
```
aleclarson: This is a small optimization, but we can avoid marking the property as "assigned" when nothing changed.

mweststrate:
Cool stuff! I'll hope to batch review / process PR's tomorrow or friday!

Op di 27 nov. 2018 om 20:55 schreef Alec Larson <notifications@github.com>:

src/proxy.js
mweststrate: Nice catch!

Issue #254, opened 2018-11-22
[RFC] Export a "produce" factory


aleclarson:
* [x] Feature request: Export a `new Immer(options)` function
  * [x] I am willing to implement this myself

## Details  

The `Immer` constructor lets you create a `produce` function with custom behavior and/or hooks.

- Promotes friendly interop between libraries using Immer
- Opens up space for niche library-specific configuration

```js
import {Immer} from 'immer'

// An example using the default values:
const {produce} = new Immer({
  useProxies: typeof Proxy !== 'undefined' && typeof Reflect !== 'undefined',
  autoFreeze: typeof process !== 'undefined' && process.env.NODE_ENV !== 'production',
  deepFreeze: false,
  hooks: {
    set(state, prop, value) {},
    delete(state, prop) {},
    copy(state) {},
  }
})

// Use `produce` like you normally would.
produce({ a: 1 }, recipe)
produce(recipe)({ a: 1 })
```

As you can see, there's plenty of room for more configuration, if and when use cases demand it.

This would supplant the need for #221.

aleclarson:
ping @mweststrate ICYMI

mweststrate:
@aleclarson I think that is a nice feature!

PR #258, opened 2018-11-30
feat: add `Immer` class

aleclarson:
Every instance of the `Immer` class provides a bound `produce` function. As explained in #254, this class makes room for advanced configuration (if we need it) and allows for interop between Immer-using libraries that have different needs.

I took the time to refactor while I was implementing this, so the diff may look
more overwhelming than it is. :)

### Summary

1. Renamed `immer.js` to `index.js` so the `Immer` class can have `immer.js`

2. Use the word "draft" in place of "proxy" in most places, because "draft" is
   a higher level of abstraction. It also makes sense to use the same semantics
   both internally and externally.

3. Moved `finalize` logic to the `Immer` class

4. Inlined the `freeze` function as only one callsite existed

5. Inlined the `createState` functions in both `proxy.js` and `es5.js`

6. Extract repeated code from `produceEs5` and `produceProxy` (which have since
   been removed) into the `Immer` class

7. The `es5.js` and `proxy.js` modules now have different exports:
    - `scopes`: an array of nested `produce` calls, where each value is an array
      of unrevoked proxies
    - `currentScope()`: shortcut for getting the last value of the `scopes` array
    - `createDraft()`: a renamed `createProxy` with the arguments swapped
    - `willFinalize()`: called right after the recipe function returns (only
      used by ES5 proxies)

8. Changed some "draft state" properties:
    - removed `hasCopy` in ES5 state (checking for truthiness of `copy` does the job)
    - renamed `proxy` to `draft` in ES5 state
    - renamed `finished` to `revoked` in ES5 state
    - renamed `proxies` to `drafts` in Proxy state
    - added `revoke` method (called by the `Immer` class)
    - added `draft` property to Proxy state
    - use null literals instead of undefined (for brevity)

9. Delay creation of `patches` and `inversePatches` arrays until the recipe
   function returns. This avoids array allocations when a rollback is performed
   by throwing.

10. Simplified `generatePatches` by removing the last two arguments, since they
    can be obtained from the `state` argument.


mweststrate:
That's fast! Quick question, any impact on build size?

Op vr 30 nov. 2018 om 16:53 schreef Alec Larson <notifications@github.com>:

mweststrate:
(if it is not too much effort, otherwise will check myself when reviewing)

Op vr 30 nov. 2018 om 17:12 schreef Michel Weststrate <mweststrate@gmail.com

aleclarson:
Note: I added lots of comments, so that probably explains the unminified bloat.

aleclarson:
Tests passing! :)

I wonder if the Coveralls bot can be fixed?

aleclarson:
I separated the diff noise out, so you can use https://github.com/mweststrate/immer/pull/258/commits/5a7d65c1e751ad4adacd1384c08dbd995c3c6ad2 for easier reviewing.

src/patches.js
```diff
@@ -1,100 +1,72 @@
...
-        else
-            generateObjectPatches(
-                state,
-                basepath,
-                patches,
-                inversePatches,
-                baseValue,
-                resultValue
-            )
+export function generatePatches(...args) {
```
aleclarson: Oh, this is probably inflated by the transpiler. Should I revert this change?

aleclarson:
Updated the `.d.ts` but not the `.js.flow` since I'm not using Flow currently.

src/immer.js
```diff
@@ -1,77 +1,168 @@
...
-            )
-        }
-    }
-
-    // prettier-ignore
-    {
-        if (typeof producer !== "function") throw new Error("if first argument is not a function, the second argument to produce should be a function")
-        if (patchListener !== undefined && typeof patchListener !== "function") throw new Error("the third argument of a producer should not be set or a function")
-    }
+        this.produce = (base, recipe, patchListener) => {
```
mweststrate: for readability you could declare this as normal method on the class, and just `this.produce = this.produce.bind(this)` in the `constructor` (also preserves the name in stack traces and such)

src/patches.js
mweststrate: Yeah I think minifier will make this pretty small anyway

src/es5.js
```diff
@@ -2,52 +2,78 @@
...
+export function createDraft(base, parent) {
+    if (isDraft(base)) throw new Error("This should never happen. Please report: https://github.com/mweststrate/immer/issues/new") // prettier-ignore
+
+    const draft = shallowCopy(base)
+    each(base, prop => {
+        Object.defineProperty(draft, "" + prop, createPropertyProxy("" + prop))
+    })
+
+    const state = {
         modified: false,
```
mweststrate: I think we get the build size down by enabling property mangling in the minifier, and just exclude the property names that are part of the public api. (Such as the public methods on an Immer instance).  Probably better to do in a separate PR though

aleclarson: The draft state becomes a public API once hooks are added. ;)

mweststrate: I just noticed. But that is only a subset of it correct? Anyway, an idea to experiment with, but not really related to this MR. Ill open fresh issue

PR #259, opened 2018-11-30
feat: add hooks to the `Immer` class

aleclarson:
Follow-up to #221  / Depends on #258 

### Summary

Hooks are passed as options to the `Immer` class constructor.

- `onAssign(state, prop, value)`
  Called for every property assigned to by a producer

- `onDelete(state, prop)`
  Called for every property deleted by a producer

- `onCopy(state)`
  Called whenever a copy becomes finalized

&nbsp;

The `state` argument is an `ImmerState` object.

https://github.com/mweststrate/immer/blob/32c45737c22ec961f6d55cce0fa0e66caadce600/src/index.d.ts#L229-L234

The `state` is guaranteed to be both modified and finalized if ever passed to a hook.

### DIFF
--- a/src/immer.js
+++ b/src/immer.js
@@ -20,7 +20,10 @@ const configDefaults = {
     autoFreeze:
         typeof process !== "undefined"
             ? process.env.NODE_ENV !== "production"
-            : verifyMinified.name === "verifyMinified"
+            : verifyMinified.name === "verifyMinified",
+    onAssign: null,
+    onDelete: null,
+    onCopy: null
 }
 
 export class Immer {
@@ -131,6 +134,12 @@ export class Immer {
         if (!state.finalized) {
             state.finalized = true
             this.finalizeTree(state.draft, path, patches, inversePatches)
+            if (this.onDelete) {
+                const {assigned} = state
+                for (const prop in assigned)
+                    assigned[prop] || this.onDelete(state, prop)
+            }
+            if (this.onCopy) this.onCopy(state)
             if (this.autoFreeze) Object.freeze(state.copy)
             if (patches) generatePatches(state, path, patches, inversePatches)
         }
@@ -148,8 +157,11 @@ export class Immer {
                 : (state.copy = shallowCopy(state.draft))
         }
         const finalizeProperty = (prop, value, parent) => {
-            // Skip unchanged properties in draft objects.
-            if (state && parent === root && is(value, state.base[prop])) return
+            if (state && parent === root) {
+                // Skip unchanged properties in draft objects.
+                if (is(value, state.base[prop])) return
+                if (this.onAssign) this.onAssign(state, prop, value)
+            }
             if (!isDraftable(value)) return
             if (!isDraft(value)) {
                 // Frozen values are already finalized.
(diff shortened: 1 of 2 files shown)


src/immer.js
```diff
@@ -131,6 +134,12 @@ export class Immer {
         if (!state.finalized) {
             state.finalized = true
             this.finalizeTree(state.draft, path, patches, inversePatches)
+            if (this.onDelete) {
+                const {assigned} = state
+                for (const prop in assigned)
+                    assigned[prop] || this.onDelete(state, prop)
```
mweststrate: I think this should be `assigned[prop] === false && this.onDelete(state, prop)` (`undefined` means nothing happened with the prop from top of my head)

src/index.d.ts
```diff
@@ -168,7 +176,13 @@ export function original<T>(value: T): T | void
 export function isDraft(value: any): boolean
 
 export class Immer {
-    constructor(config: {useProxies?: boolean; autoFreeze?: boolean})
+    constructor(config: {
+        useProxies?: boolean
+        autoFreeze?: boolean
+        onAssign?: <T>(state: ImmerState<T>, prop: keyof T, value: any) => void
```
mweststrate: Not sure if `<T>` adds much value here, since the callback is called with different types of T and a single T propbably doesn't describe any object in the tree.  On the other hand it isn't in the way either.

src/index.d.ts
```diff
@@ -211,3 +225,10 @@ export class Immer {
      */
     setUseProxies(useProxies: boolean): void
 }
+
+export interface ImmerState<T = any> {
+    parent?: ImmerState
+    base: T
+    copy: T
+    assigned: {[prop: string]: boolean}
```
mweststrate: Not 100% sure about the `assigned` field. On one hand it feels like leaking internal details, that can also be determined by the other hooks. On the other hand, it might be super handy, so probably we can leave it.

aleclarson:
> P.S.: random idea, would it be cool to in the future create a mini-immer that has no patches, hooks, or ES5 support?

Yeah, that sounds pretty nice. 👍 Especially useful if you're only using Immer for reducers.

src/immer.js
aleclarson: The `in` operator protects against that. :)

mweststrate:
Lol, check.

Op do 13 dec. 2018 om 15:55 schreef Alec Larson <notifications@github.com>:

src/index.d.ts
aleclarson: I don't remember why I added that, so I'll just remove it. 👍

src/index.d.ts
aleclarson: @mweststrate Yeah, it's just nice to reuse the data instead of recomputing it manually. :)

mweststrate: I can imagine, let's leave it

src/immer.js
```diff
@@ -20,7 +20,10 @@ const configDefaults = {
     autoFreeze:
         typeof process !== "undefined"
             ? process.env.NODE_ENV !== "production"
-            : verifyMinified.name === "verifyMinified"
+            : verifyMinified.name === "verifyMinified",
+    onAssign: null,
+    onDelete: null,
+    onCopy: null
```
aleclarson: Should I remove these lines for brevity?

aleclarson:
Merging this into the other PR.

PR #274, opened 2018-12-18
[TS] Allow extraArgs for curried producer with initial state

geirsagberg:
After upgrading to TypeScript 3.2.2, I started getting errors when using immer for reducers with initial state:

image

It worked if I used the curried producer without initial state, but not with initial state. So I tried adding `extraArgs` to the curried producer with initial state, and it started working again.


aleclarson:
Yeah, I caught this one yesterday, but haven't pushed the fix (which is slightly different than this). I'll have it released in v1.9.3. Thanks for contributing!

Issue #246, opened 2018-11-09
Idea: Ignore return value of producer


mweststrate:
Currently the return value (if any) matters a lot, to be able to replace the current state. So there are quite some rules

1. Return `undefined` (or no `return` at all), will assume the `draft` is the state to be returned
2. Return any value (except `draft` or `undefined`), will replace the state with the returned value

This cause a few tricky cases:
* Because of 1, we cannot replace the current state with `undefined`, by simple returning `undefined`. For that reason, `nothing` has to be returned
* Because of 2, it is easy to accidentally replace the state, for example in `produce({ x: 1}, draft => draft.x = 2)` will produce `2` as the next state, not `{ x: 2 }`. (Since assignment is statements return their right hand value, the result of the function is `2`, not `undefined`. To avoid this, we have to write either `draft => { draft.x = 2 }` or `draft => void draft.x = 2`.

We could design the API differently, and always by default ignore the return value of a producer, regardless what the return value is. This fixes point 2, where we accidentally replace the state. 

To be able to replace the state, we could have a dedicated internal indicating that we intend the replacement: `return replace(3)`. This also solves 1), as `undefined` has no longer to be treated specially, and we can just `return replace(undefined)`. 

Note that we could even detect a `replace` call without `return` and warn the user. 

Beyond that, we can possible warn about return a value, but not updating the draft (probably replace was intended)

So, in summary:

```javascript
import { produce, replace } from "immer"

produce({ x: 1 }, draft => draft.x = 2) 
// returns { x: 2 }

produce({x : 1}, draft => replace(2))
// returns 2

produce({x : 1}, draft => replace(undefined))
// returns undefined

produce({ x: 1 }, draft => {
   return replace(3)
})
// returns 3

produce({ x: 1 }, draft => {
   draft.x = 2
   return 3
})
// returns { x: 2 }
// The '3' is ignored entirely (and we can't really warn about it)
```

Alternatively we could warn about non-`undefined` return values (fixing the last case) and forcing consumers to either use `d => { ...  }` or `d => void ... ` or `d => replace(...)` in producers.

_Edit: removed the warnings_

aleclarson:
IMO: Warn when `draft` is mutated but _not_ returned. Then whenever a producer returns `undefined`, replace the `draft` with `undefined`.

aleclarson:
[...]

PR #279, opened 2018-12-18
feat: always use the return value

aleclarson:
__Provoked by:__ https://github.com/mweststrate/immer/issues/246

__Goal:__ Simplify the mental model of producers

__Hypothesis:__ When `produce` returns exactly what its recipe function returns (apart from proxies, of course), complexity is reduced for developers (and TypeScript).

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -1,8 +1,3 @@
-export const NOTHING =
-    typeof Symbol !== "undefined"
-        ? Symbol("immer-nothing")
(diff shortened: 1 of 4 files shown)


oriSomething:
Isn't it a major breaking change?

aleclarson:
> Isn't it a major breaking change?

Yes, the commit message says that.

mweststrate:
I think this one can be closed, given conclusion in #246 to keep things as-is

PR #282, opened 2018-12-29
feat: passing drafts as the base state (or in the base state)

aleclarson:
Improve support for nested producers by allowing drafts to be passed to `produce` as the base state (or in the base state).

The main commit: e3fc824a17791823088ea23f7631248cd0187173

### Details

1. Added `state.scope` for knowing when a draft is owned by the current
  producer or not.

1. Objects included in the return value of a nested producer are
never auto-freezed, because they may contain drafts from a parent producer.

1. Drafts are always finalized by the producer that created them.

1. Now that `state.base` can be a draft, we must ensure `state.finalized`
  is false before creating drafts in the `get` handler. This only affects ES2015+ proxies.


aleclarson:
**Question about detail number 2:** Is this the best behavior? Or would it be better to pre-emptively clone and swap out any parent-owned drafts found in the returned state?

mweststrate:
> @mweststrate Is this the best behavior? Or would it be better to pre-emptively clone and swap out any parent-owned drafts found in the returned state?

Yeah, I think this is the best behavior for now, "things are frozen when outer producer finishes" is a fine semantic I think. Might actually solve some issues we saw in the past by higher order reducer that secretly extended state with their own meta data.

mweststrate:
Looking good @aleclarson! I can't think of any edge cases atm that aren't cover. You neither I guess 😎?. 

I think we can release this as minor version?

aleclarson:
@mweststrate Awesome! I'm working on non-enumerable properties and class support right now. 🤓 Should we ship them together?

mweststrate:
Releasing is cheap. So preferably not, releasing separately makes it easier
to pinpoint regression bugs

Op vr 4 jan. 2019 om 16:44 schreef Alec Larson <notifications@github.com>:

PR #305, opened 2019-01-30
feat: support async recipes

aleclarson:
Drafts won't be revoked until the returned promise is fulfilled or rejected.

The readme needs to be updated to reflect this. Anyone willing to contribute here?

As suggested by @Yurickh in https://github.com/mweststrate/immer/issues/302#issuecomment-457839538

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -833,57 +833,61 @@ function runBaseTest(name, useProxies, autoFreeze, useListener) {
             })
         })
 
+        // TODO: rewrite tests with async/await once node 6 support is dropped
         describe("async recipe function", () => {
-            it("can modify the draft", async () => {
+            it("can modify the draft", () => {
                 const base = {a: 0, b: 0}
-                const res = await produce(base, async d => {
+                return produce(base, d => {
                     d.a = 1
-                    await Promise.resolve()
-                    d.b = 1
+                    return Promise.resolve().then(() => {
+                        d.b = 1
+                    })
+                }).then(res => {
+                    expect(res).not.toBe(base)
+                    expect(res).toEqual({a: 1, b: 1})
                 })
-                expect(res).not.toBe(base)
-                expect(res).toEqual({a: 1, b: 1})
             })
 
-            it("works with rejected promises", async () => {
+            it("works with rejected promises", () => {
                 let draft
                 const base = {a: 0, b: 0}
                 const err = new Error("passed")
-                try {
-                    await produce(base, async d => {
-                        draft = d
-                        draft.b = 1
-                        await Promise.reject(err)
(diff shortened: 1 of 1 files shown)


mweststrate:
questions that just popped up in my mind: does this affect `isDraft` / `original` / recursive produce calls if they are invoked after the first `await`?

aleclarson:
Note: When using an async recipe function, you're free to "hoist" drafts out of their recipe function, since they'll definitely be finalized when the promise is resolved. But it remains unsafe to do the same with a synchronous recipe function.

aleclarson:
Note: With async recipe functions, it remains unsafe to "hoist" drafts out of the recipe, since they can be revoked at any time.

mweststrate:
also note that in that sense it also can' t be used to solve the lowlevel
proposal in 302; as finalizing would always happen on at least one tick
later; when the promise settles

Op do 31 jan. 2019 om 16:37 schreef Alec Larson <notifications@github.com>:

aleclarson:
> as finalizing would always happen on at least one tick later; when the promise settles

I don't think so, because you're not *required* to use promises when deferring the finalize phase. I've separated that logic into the `processResult` method. **But,** it would only be safe to call `finishDraft` on the base draft; never on a child draft (at least with the current implementation).

mweststrate:
ah, that is awesome @aleclarson!

aleclarson:
Note: With the current implementation, you wouldn't be able to call `finishDraft` more than once on any draft related to a single `createDraft` call, because they all share the same `ImmerScope` instance.

aleclarson:
The issue mentioned in https://github.com/mweststrate/immer/pull/305#issuecomment-459385591 has been fixed. 🎉

aleclarson:
@mweststrate Node 6 doesn't support async/await, so the tests are failing. Maybe we should drop Node 6 support? :P

Yurickh:
> @mweststrate Node 6 doesn't support async/await, so the tests are failing. Maybe we should drop Node 6 support? :P

THAT would be a breaking change after all haha

aleclarson:
@Yurickh Yup! 😆 I decided against that silly idea.

We can revert https://github.com/mweststrate/immer/pull/305/commits/5a99c88cc72e2000d2ec9268a454ede3dc86cea3 when the time comes!

aleclarson:
@mweststrate Were you having trouble editing this PR? Or just want to avoid adding another git remote?

mweststrate:
The latter, wasn't sure yet whether to use one or two branches. But changes
are so small so far I'll keep it as one

Op za 2 feb. 2019 14:34 schreef Alec Larson <notifications@github.com:

Issue #247, opened 2018-11-09
Use terser instead of uglify-es


aleclarson:
@gaearon has been tweeting about `uglify-es` causing issues for CRA users. Considering `uglify-es` is unmaintained, we should use [terser](https://www.npmjs.com/package/terser) instead (a fork of `uglify-es` that is actively maintained).

mweststrate:
Agreed

Op vr 9 nov. 2018 om 16:48 schreef Alec Larson <notifications@github.com>:

aleclarson:
I'm not using the minified version of Immer, so I cannot set aside time to do this. 😢

dwelle:
@mweststrate what's [this line](https://github.com/mweststrate/immer/blob/899b53f519b304652461348c06ea0a5e210cf2e7/rollup.config.js#L54) about? The `rollup-plugin-uglify` doesn't take any 2nd arg.

mweststrate:
@dwelle no clue, probably copied it from somewhere else, but it seems to parameterize the minifier that should be used :)

aleclarson:
We could stop bundling Immer altogether.

Bundling is the job of the application, not the library.

Web apps can use https://packd.now.sh/immer@1.8.0 or something like Webpack.

Node apps can use something like https://github.com/zeit/ncc

prithvi16:
@mweststrate are there any contributing guidelines or local development setup instructions? I looked for them in README and wiki.

mweststrate:
Bundling is pretty essential for UMD builds, for other builds it can be left out indeed

@prithvi16 no, no clear guidelines atm :), setup is pretty straight forward, make sure to use `yarn`  and that your IDE respects prettier, and that should be it. Discuss large changes upfront in an issue to avoid wasting effort

aleclarson:
@mweststrate There's https://unpkg.com/ for UMD bundles.

mweststrate:
unpkg is a CDN, not a bundler, so if you put an UMD bundle in your package,
then you can CDN UMD builds. Otherwise you have useless require / import
statements in your browser

aleclarson:
Just realized Packd creates a UMD bundle, so I think we can stop using Rollup locally. :)

aleclarson:
Note: https://bundle.run should be used over https://packd.now.sh

aleclarson:
Alternatively, we could use https://github.com/developit/microbundle (if you still believe Immer should be bundled before publishing).

mweststrate:
[...]

PR #306, opened 2019-01-31
chore: replace rollup with bili

egoist:
This is a follow-up PR to #247

Changes:

- Output ESM bundle to `dist/immer.mjs`
- Using buble instead of babel-preset-env for smaller output


egoist:
Somehow the node 8 job failed

aleclarson:
> hmmm Travis is really slow 😅 CircleCI can finish the same tests in 30 seconds

Oh so *that's* why people use CircleCI. 😆 

Now the Node 6 build is saying:
```
$ /home/travis/build/mweststrate/immer/node_modules/.bin/bili
(node:5727) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): SyntaxError: Unexpected identifier
```
Yet it still passes. 😳

aleclarson:
Aww, just realized that Bublé does not support `static` class properties, which are used in #305. 😢

egoist:
@aleclarson you can still add Babel plugins, Bublé is just used to replace babel-preset-env

aleclarson:
@egoist Interesting! I'm surprised that actually improves performance, since you're then parsing the AST twice (and still using Babel's code generator) if I understand correctly.

egoist:
The unhandled promise rejection is gone now.

Issue #308, opened 2019-02-02
Remove support for `this = draft` in recipes


mweststrate:
Currently, `this` is always equal to the draft inside recipes. I suggest to remove that. It has a few benefits:

1. simpler typings
2. simpler docs
3. I didn't see anyone use it
3. it makes pattern like this possible (note that we should make sure that the curried version does preserve `this`):

```
const context = {
  foo: 1,
  bar: produce(function(draft) {
    draft.foo = this.foo
  })
}
```

scriby:
+1 to this

mweststrate:
Done in the `immer3` branch.

PR #309, opened 2019-02-03
immer 3.0

mweststrate:
* [x] Implement #308: no `this`
* [ ] ~Implement #246, ignore return value, use `return replace(value)` for replacement~
* [x] Merge #232: improved flow typings
* [ ] ~Improve minification by using property mangling?~
* [x] Drop support for Node 6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ script:
     - yarn coveralls
     - yarn test:flow
     - yarn test:dts
-    - yarn add typescript@3.3 -D && yarn test:dts
+    - yarn add typescript@latest -D && yarn test:dts
 jobs:
     include:
         - stage: deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ script:
     - yarn coveralls
     - yarn test:flow
     - yarn test:dts
-    - yarn add typescript@latest -D && yarn test:dts
 jobs:
     include:
         - stage: deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-    - "6"
     - "8"
     - "10"
     - "node"
--- a/readme.md
+++ b/readme.md
@@ -8,7 +8,7 @@ _Create the next immutable state tree by simply modifying the current tree_
 
 Winner of the "breakthrough of the year" [open source award](https://osawards.com/react/) in 2019
 
-### [Release notes](https://github.com/mweststrate/immer/releases)
+### [Release notes](https://github.com/immerjs/immer/releases)
 
 Did Immer make a difference to your project? Consider buying me a coffee!<br/><a href="https://www.buymeacoffee.com/mweststrate" target="_blank"><img src="https://www.buymeacoffee.com/assets/img/custom_images/orange_img.png" alt="Buy Me A Coffee" style="height: auto !important;width: auto !important;" ></a>
 
--- a/readme.md
+++ b/readme.md
@@ -635,28 +635,9 @@ const newState = increment(state, 2)
 // `newState.x` cannot be modified here
 ```
 
-**Note:** Immer v1.9+ supports Typescript v3.1+ only. **Note:** Immer v2.2+ supports Typescript v3.4+ only.
+**Note:** Immer v1.9+ supports TypeScript v3.1+ only.
 
-## Using `this`
-
-The recipe will be always invoked with the `draft` as `this` context.
-
-This means that the following constructions are also valid:
-
-```javascript
-const base = {counter: 0}
-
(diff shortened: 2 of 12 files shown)


alitaheri:
I agree with @aleclarson on this. I also agree that the void trick should be removed from the docs.

But I disagree that we should always return the draft. because that would confuse someone looking at it for the fist time.

right now, what we have is perfect. just remove void trick. I have the following justifications:

1. We shouldn't extend the api beyond the scope for no real gain
2. This library doesn't represent traditional reducers, so it shouldn't be confused with them, we're recording mutations and building a new state from it, not *reducing* state with mutation to get a new state.
3. we should encourage readable code. the `void (...)` pattern is not readable for someone who doesn't know why it's there. because they should know what `void` does, keep in mind that `=> expr` returns the evaluated value and why it's there in the first place (returned values the callback mean something special)
4. We need a way to support changing the end value *in a composeable* way! `replace(value)` specializes this functionality. so a function from another library, or a shared function must *be aware of* immer. which is a framework smell. I don't like the `nothing` solution either, but that's on javascript not immer.

mweststrate:
@alitaheri would you mind moving that comment to #246? :)

aleclarson:
I'll review this today, too. (New York timezone)

Can you reword the commit messages to fit the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) style?

aleclarson:
Also, I'm a little worried about the changes to the flow types, considering [this comment](https://github.com/immerjs/immer/pull/232#issuecomment-437453053). Maybe it would be best to leave the flow types alone, or remove them entirely?

migueloller:
@aleclarson, I agree with your sentiment. The general intent of the changes I proposed are in the right direction, I think, but unless a newer version of Flow fixed a few things there were some issues with those Flow types.

This is nothing that can't be tested, though. So if all the Flow tests are passing, and we should make sure they're not false positives, it should be alright. Perhaps adding some more Flow tests could help as well?

If this is a change that you guys really want to get in for v3 maybe I can set aside some time to look into it this week.

mweststrate:
@migueloller, ah misunderstood the latest commits, thought that they addressed the open issues. Will revert this change (we can also merge it in later).

I peeked in more detail at your changes, at least they are partially incorrect: `produce` (uncurried) always produces `S`, not `R`! Same for the curried function, that also produces `S`. The recipe itself can indeed, in both cases, either return `S | void`.

mweststrate:
@aleclarson sorry, I really should make that a habit! Rebasing...

mweststrate:
eh lol, I'll try again

mweststrate:
Note: build is currently not being executed due to the repo transfer. It didn't sync to travis yet....

aleclarson:
LGTM! 👍 

I simplified the `IProduce` type tests ([see here](https://github.com/immerjs/immer/pull/309/commits/cc5f52bfcb504b6da0527186117166e5028c10fe)). :)

.travis.yml
```diff
@@ -18,7 +17,7 @@ script:
     - yarn coveralls
     - yarn test:flow
     - yarn test:dts
-    - yarn add typescript@3.3 -D && yarn test:dts
+    - yarn add typescript@latest -D && yarn test:dts
```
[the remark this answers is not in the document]
mweststrate: I think I'll revert it to 3.4, some hardcoded versions. TS is quite aggressive with their latest tag, and having random build failures is annoying. It is not the CI's responsibility to test compatible with random TS versions and gives us heads ups. The point is to make sure we didn't break anything _ourselves_

mweststrate:
@aleclarson. Lol, thanks for doing the obvious thing :)

readme.md
```diff
@@ -734,10 +735,18 @@ Most important observation:
 
 ## Migration
 
+**Immer 2.\* -> 3.0**
+
+TODO
```
aleclarson: Whoops ;)

aleclarson: I cancelled the deploy for you 👍

mweststrate:
Lol, something isn't working as supossed to. Hopefully release works fine
on next build, otherwise, could you trigger the 3.0 release?

aleclarson:
@mweststrate Did you see my review comment about the migration guide?

mweststrate:
@aleclarson ehh nope. I have to leave currently. Would you mind triggering a 3.0 release, and updating migration guide if needed? I didn't figure out how to trigger a major release with semantic-release yet.

aleclarson:
The `BREAKING CHANGE` part goes in the footer 😛

mweststrate:
Argh

Op wo 17 apr. 2019 17:17 schreef Alec Larson <notifications@github.com>:

PR #320, opened 2019-02-25
Array patches were optimized

ikabirov:
Fix for https://github.com/mweststrate/immer/issues/319

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -76,19 +76,6 @@ describe("applyPatches", () => {
             applyPatches({}, [patch])
         }).toThrowError(/^Cannot apply patch, path doesn't resolve:/)
     })
-    it("throws when a patch tries to splice an array", () => {
-        // Pop is ok
-        expect(() => {
-            const patch = {op: "remove", path: [0]}
-            applyPatches([1], [patch])
-        }).not.toThrowError()
-
-        // Splice is unsupported
-        expect(() => {
-            const patch = {op: "remove", path: [0]}
-            applyPatches([1, 2], [patch])
-        }).toThrowError(/^Only the last index of an array can be removed/)
-    })
 })
 
 describe("simple assignment - 1", () => {
@@ -108,6 +95,16 @@ describe("simple assignment - 2", () => {
             d.x.y++
         },
         [{op: "replace", path: ["x", "y"], value: 5}]
+    )
+})
+
+describe("simple assignment - 3", () => {
+    runPatchTest(
+        {x: [{y: 4}]},
+        d => {
+            d.x[0].y++
+        },
+        [{op: "replace", path: ["x", 0, "y"], value: 5}]
     )
 })
 
@@ -198,11 +195,20 @@ describe("arrays - prepend", () => {
         d => {
             d.x.unshift(4)
         },
+        [{op: "add", path: ["x", 0], value: 4}]
+    )
+})
+
+describe("arrays - multiple prepend", () => {
+    runPatchTest(
+        {x: [1, 2, 3]},
+        d => {
+            d.x.unshift(4)
+            d.x.unshift(5)
+        },
         [
-            {op: "replace", path: ["x", 0], value: 4},
-            {op: "replace", path: ["x", 1], value: 1},
-            {op: "replace", path: ["x", 2], value: 2},
-            {op: "add", path: ["x", 3], value: 3}
+            {op: "add", path: ["x", 0], value: 5},
+            {op: "add", path: ["x", 1], value: 4}
         ]
     )
 })
@@ -213,8 +219,35 @@ describe("arrays - splice middle", () => {
         d => {
             d.x.splice(1, 1)
         },
+        [{op: "remove", path: ["x", 1]}]
+    )
+})
+
+describe("arrays - multiple splice", () => {
+    runPatchTest(
+        [0, 1, 2, 3, 4, 5, 0],
+        d => {
+            d.splice(4, 2, 3)
+            d.splice(1, 2, 3)
+        },
         [
-            {op: "replace", path: ["x", 1], value: 3},
+            {op: "replace", path: [1], value: 3},
+            {op: "replace", path: [2], value: 3},
+            {op: "remove", path: [5]},
+            {op: "remove", path: [4]}
+        ]
+    )
+})
+
+describe("arrays - modify and shrink", () => {
+    runPatchTest(
+        {x: [1, 2, 3]},
+        d => {
+            d.x[0] = 4
+            d.x.length = 2
+        },
+        [
+            {op: "replace", path: ["x", 0], value: 4},
             {op: "replace", path: ["x", "length"], value: 2}
         ]
     )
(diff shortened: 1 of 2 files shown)


src/patches.js
```diff
@@ -87,15 +129,19 @@ export function applyPatches(draft, patches) {
...
                 case "replace":
-                case "add":
-                    // TODO: add support is not extensive, it does not support insertion or `-` atm!
                     base[key] = patch.value
                     break
+                case "add":
+                    if (Array.isArray(base)) {
+                        base.splice(key, 0, patch.value)
+                    } else {
+                        // TODO: add support is not extensive, it does not support insertion or `-` atm!
```
aleclarson: Delete this comment

src/patches.js
```diff
@@ -87,15 +129,19 @@ export function applyPatches(draft, patches) {
             const key = path[path.length - 1]
             switch (patch.op) {
                 case "replace":
-                case "add":
-                    // TODO: add support is not extensive, it does not support insertion or `-` atm!
                     base[key] = patch.value
                     break
+                case "add":
+                    if (Array.isArray(base)) {
+                        base.splice(key, 0, patch.value)
```
aleclarson: Above this line, add `// TODO: support "foo/-" paths for appending to an array`

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
-            path: basePath.concat("length"),
-            value: base.length
-        })
     }
+    start += replaceCount
 
-    // ...or did it shrink?
-    else if (minLength < base.length) {
+    const useRemove = start != base.length
+    for (let j = 0; j < delta; ++j) {
```
aleclarson: There's no need to use `j` instead of `i` here.

__tests__/patch.js
```diff
@@ -76,7 +76,7 @@ describe("applyPatches", () => {
             applyPatches({}, [patch])
         }).toThrowError(/^Cannot apply patch, path doesn't resolve:/)
     })
-    it("throws when a patch tries to splice an array", () => {
+    it("not throws when a patch tries to splice an array", () => {
```
aleclarson: Replace with `it("supports array splice and pop"`

__tests__/patch.js
```diff
@@ -87,7 +87,7 @@ describe("applyPatches", () => {
         expect(() => {
             const patch = {op: "remove", path: [0]}
             applyPatches([1, 2], [patch])
-        }).toThrowError(/^Only the last index of an array can be removed/)
+        }).not.toThrowError()
```
aleclarson: Remove both `expect(() => {` and `}).not.toThrowError()`.

Test that the result of `applyPatches` equals `[2]`.

And remove both comments in this test.

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
-    const minLength = Math.min(base.length, copy.length)
+function _generateArrayPatches(
+    base,
+    copy,
+    assigned,
+    basePath,
+    patches,
+    inversePatches
+) {
+    if (copy.length < base.length) {
```
aleclarson: How does this branch work? You swap `copy` with `base` and `patches` with `inversePatches`. Why?

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
-    // Look for replaced indices.
-    for (let i = 0; i < minLength; i++) {
-        if (assigned[i] && base[i] !== copy[i]) {
-            const path = basePath.concat(i)
-            patches.push({op: "replace", path, value: copy[i]})
-            inversePatches.push({op: "replace", path, value: base[i]})
-        }
+    const delta = copy.length - base.length
+
+    let start = 0
```
aleclarson: This needs a comment. What is this the `start` of?

Something like `// Find the first changed index.`

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
+            path,
+            value: copy[start + j]
         })
-        for (let i = minLength; i < base.length; i++) {
-            inversePatches.push({
-                op: "add",
-                path: basePath.concat(i),
-                value: base[i]
+        if (useRemove) {
+            inversePatches.unshift({
```
aleclarson: Why are you using `unshift` here? And why is the inverse patch conditional when its associated patch is not?

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
+                op: "remove",
+                path
             })
         }
     }
+
+    if (!useRemove) {
+        inversePatches.push({
```
aleclarson: Every inverse patch needs an associated patch, but this doesn't seem to have one. Why?

Also, an inverse patch that replaces the `length` is useless. What about the values that were truncated?

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
+function _generateArrayPatches(
+    base,
+    copy,
+    assigned,
+    basePath,
+    patches,
+    inversePatches
+) {
+    if (copy.length < base.length) {
+        return _generateArrayPatches(
```
aleclarson: Not a fan of the underscore. Since you're just swapping arguments, maybe do this instead:
```js
;[copy, base] = [base, copy]
;[patches, inversePatches] = [inversePatches, patches]
```

src/patches.js
```diff
@@ -6,50 +6,92 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
+
+    if (!useRemove) {
+        inversePatches.push({
+            op: "replace",
+            path: basePath.concat(["length"]),
+            value: base.length
+        })
+    }
+
+    return true
```
aleclarson: Remove this line

src/patches.js
ikabirov: Swap `copy` with `base` and `patches` with `inversePatches` returns the same result, but it's normalize arrays and `delta` always be positive or 0

src/patches.js
ikabirov: Remove operations shifts indices, so we add remove operations in reverse order. Inverse patch ignores when we can use replace 'length' property instead of multiple remove operations

src/patches.js
ikabirov: It'a alternative for multiple 'remove' operation. So we use only one of them

Set length for arrays is standard practice for JS. Truncated items will be cleared
```
var a = [1,2,3];
a.length = 1;
a.length = 2;
console.log(a); // [1, _empty_]
```

src/patches.js
```diff
@@ -6,50 +6,89 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
-            path: basePath.concat("length"),
-            value: base.length
-        })
     }
+    start += replaceCount
 
-    // ...or did it shrink?
-    else if (minLength < base.length) {
+    // If add items to end, we can use replace 'length' property
+    // instead of multiple remove operations
```
aleclarson: Use this comment:

```js
// For "add" patches that extend the array, we can use a single "replace" patch (on the
// `length` property) as the inverse patch, instead of multiple "remove" patches.
```

src/patches.js
```diff
@@ -6,50 +6,89 @@ export function generatePatches(state, basePath, patches, inversePatches) {
...
-        }
+    const delta = copy.length - base.length
+
+    // Find the first changed index.
+    let start = 0
+    while (base[start] === copy[start] && start < base.length) {
+        ++start
+    }
+
+    // Find the first changed index from right
```
aleclarson: Use this comment:
```js
// Find the last changed index.
```

src/patches.js
ikabirov: No. Add operations should be applied in ascend order, remove operations in descend

aleclarson: Right, you would do it like this:
```js
for (let i = delta - 1; i >= 0; --i) {
  const path = basePath.concat([start + i])
  patches[i] = {
    op: "add",
    path,
    value: copy[start + i]
  }
  if (useRemove) {
    inversePatches.push({
      op: "remove",
      path
    })
  }
}
```

src/patches.js
ikabirov: This behavior was implemented by @mweststrate, I just kept it
```
describe("arrays - push multiple", () => {
    runPatchTest(
        {x: [1, 2, 3]},
        d => {
            d.x.push(4, 5)
        },
        [
            {op: "add", path: ["x", 3], value: 4},
            {op: "add", path: ["x", 4], value: 5}
        ],
        [{op: "replace", path: ["x", "length"], value: 3}]
    )
})
```

aleclarson: Ah, I missed that test! 👍

aleclarson:
Fantastic work, @ikabirov! 👏 

I did a little refactoring, and the tests are still passing. LGTM! 🎉

PR #325, opened 2019-03-07
TypeScript 3.4.0 support

aleclarson:
Noticed an issue with the `Immutable<T>` type while testing out `typescript@3.4.0-dev.20190306`.

This PR will also track other Immer-related issues with TypeScript 3.4.0 until 3.4.0 is officially released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ script:
     - yarn coveralls
     - yarn test:flow
     - yarn test:dts
-    - yarn add typescript@3.3 -D && yarn test:dts
+    - yarn add typescript@latest -D && yarn test:dts
 jobs:
     include:
         - stage: deploy
--- a/__tests__/immutable.ts
+++ b/__tests__/immutable.ts
@@ -14,19 +14,19 @@ declare const exactType: <Actual, Expected>(
 // array in tuple
 {
     let val = {} as Immutable<[string[], 1]>
-    exactType(val, {} as [ReadonlyArray<string>, 1])
+    exactType(val, {} as readonly [ReadonlyArray<string>, 1])
 }
 
 // tuple in array
 {
     let val = {} as Immutable<[string, 1][]>
-    exactType(val, {} as ReadonlyArray<[string, 1]>)
+    exactType(val, {} as ReadonlyArray<readonly [string, 1]>)
 }
 
 // tuple in tuple
 {
     let val = {} as Immutable<[[string, 1], 1]>
-    exactType(val, {} as [[string, 1], 1])
+    exactType(val, {} as readonly [readonly [string, 1], 1])
 }
 
 // array in array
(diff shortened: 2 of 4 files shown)


aleclarson:
We need a way to run TypeScript tests based on the installed version, since there are breaking changes between TypeScript 3.4.1 and 3.3.4000. Here are the solutions I can think of:

1. During each CI build, apply a git patch to our tests and install the TypeScript version associated with said patch. This could be done for multiple TypeScript versions with breaking changes.

2. Drop support for TypeScript 3.3.x and under

3. Find another way... 😅 

Thoughts, @mweststrate?

aleclarson:
Oh, I forgot about `typesVersions` introduced in TypeScript 3.1:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions

aleclarson:
This PR is dependent on https://github.com/prettier/prettier/issues/5977 being fixed.

aleclarson:
> During each CI build, apply a git patch to our tests and install the TypeScript version associated with said patch. This could be done for multiple TypeScript versions with breaking changes.

It would be nice if we could only run the TypeScript tests when we know the typedefs have changed, but I don't think anyone has solved that yet. Otherwise, CI builds are slowed down by needing to install multiple versions of TypeScript to ensure safe upgrades / backward-compatible changes.

jineshshah36:
I'm getting `Type instantiation is excessively deep and possibly infinite.` pretty much in every producer on 3.4. I found that making the following change fixed it:

```ts
// old
// export type Draft<T> = T extends never[]
//   ? T
//   : T extends ReadonlyArray<any>
//   ? T[number][] extends T
//     ? DraftArray<T>
//     : DraftTuple<T>
//   : T extends AtomicObject
//   ? T
//   : T extends object
//   ? { -readonly [P in keyof T]: Draft<T[P]> }
//   : T

// new
export type Draft<T> = { -readonly [K in keyof T]: Draft<T[K]> }
```
Since 3.4 improves support for readonly arrays and tuples, this seems to work without breaking anything.

brunolemos:
Please release a new version with typescript 3.4 support, the latest stable vscode already comes with this version.

I'm also getting `Type instantiation is excessively deep and possibly infinite.` on every `immer` producer. Using typescript `3.4.3`.

jineshshah36:
@brunolemos that is not a valid workaround. At a minimum, you need something like the following in order to correctly type the draft:

```ts
export type Draft<T> = T extends AtomicObject
  ? T
  : T extends object
  ? { -readonly [K in keyof T]: Draft<T[K]> }
  : T
```

The reason is if T has any readonly properties, they need to be erased

mweststrate:
Lol, it was already in there. Awesome test setup btw @aleclarson!

mweststrate:
closing in favor of #348 (which was based branched from this PR)

PR #334, opened 2019-03-14
I think RN 0.59 no longer needs Proxy polyfill on Android

Noitidart:
With React Native 0.59 release - they upgraded the Android JSC so I think this polyfill is no longer needed - https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md

I might be wrong here, can someone else confirm please?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -505,7 +505,7 @@ Immer automatically freezes any state trees that are modified using `produce`. T
 
 ## Immer on older JavaScript environments?
 
-By default `produce` tries to use proxies for optimal performance. However, on older JavaScript engines `Proxy` is not available. For example, when running Microsoft Internet Explorer or React Native on Android. In such cases, Immer will fallback to an ES5 compatible implementation which works identical, but is a bit slower.
+By default `produce` tries to use proxies for optimal performance. However, on older JavaScript engines `Proxy` is not available. For example, when running Microsoft Internet Explorer or React Native (< v0.59) on Android. In such cases, Immer will fallback to an ES5 compatible implementation which works identical, but is a bit slower.
 
 ## Importing immer
 


mweststrate:
Could you test it? simply use `setUseProxies(true)`, and proxy
implementation is forced, so if producers than still pass a basic test, it
works

Noitidart:
Roger that! Will test it and get back to you soon! Thanks!

Noitidart:
Awesome I think it works! I did this:

```
import produce, { setUseProxies } from 'immer';

setUseProxies(true);

const baseState = {
  hi: true
};

const nextState = produce(baseState, draft => {
  draft.bye = true;
});

console.log('baseState === nextState:', baseState === nextState);
```

And it correctly logged `false`!

I also tried:

```

const baseState = {
  hi: true
};

const nextState = produce(baseState, draft => {
  draft.hi = false;
  draft.hi = true;
});

console.log('baseState === nextState:', baseState === nextState);
```

And this also correctly logged `true`.

Noitidart:
I think I'm getting a bug, but this is giving me same with proxies on and off:

```
const baseState = {
  hi: true
};

const nextState = produce(baseState, draft => {
  draft.hi = false;
  draft.hi = true;
});

console.log('baseState === nextState:', baseState === nextState);
```

`nextState` should be the same here. But its giving me `false`. :(

aleclarson:
@Noitidart That test should pass with `setUseProxies(false)` and fail with `setUseProxies(true)`,

https://github.com/mweststrate/immer/blob/8b33b9882e3c0981e04ee3cd1edd7a9d774612e7/__tests__/base.js#L124-L136

Noitidart:
Oh interesting @aleclarson - I had done `setUseProxies(true)`, so this means that proxy support is not yet there in RN on Android?

mweststrate:
@Noitidart the fact that you can run a producer, with `setProxies(true)` means that proxy support is enabled.

PR #348, opened 2019-04-10
Ts 3.4 inference improvements

mweststrate:
Improved type inference a bit:

* Required version is upped to 3.4
* Draft and Immutable have become simpler, as suggest in #325, which is possible since 3.4
* Type inference for curried `produce` has improved. Especially the rest parameters are now better typed and has less edge cases (before at random places `| boolean` was added to the type inference, and rest arguments remained optional`
  * As a result `produce<State>(fn)` no longer works, however, `produce((x: Draft<State>))` will lead to the correct inference

@aleclarson feel free to shoot at it! I think especially for currying type inference is now better and a bit more predictable, but this change is also kinda breaking on how to type producers

### DIFF
--- a/__tests__/produce.ts
+++ b/__tests__/produce.ts
@@ -79,10 +79,11 @@ it("can infer state type from default state", () => {
 
 it("can infer state type from recipe function", () => {
     type State = {readonly a: string} | {readonly b: string}
-    type Recipe = (base: State) => State
+    type Recipe = (base: State, n: number) => State
 
-    let foo = produce((draft: Draft<State>) => {})
+    let foo = produce((draft: Draft<State>, number: number) => {})
     exactType(foo, {} as Recipe)
+    const state: State = foo({} as State, 3)
 })
 
 it("can infer state type from recipe function with arguments", () => {
@@ -295,4 +296,69 @@ it("works with generic parameters", () => {
     let val: {readonly a: ReadonlyArray<number>} = 0 as any
     let arr: ReadonlyArray<typeof val> = 0 as any
     insert(arr, 0, val)
+})
+
+it("can work with non-readonly base types", () => {
+    const state = {
+        price: 10,
+        todos: [{
+            title: "test",
+            done: false
+        }]
+    }
+    type State = typeof state
+
+    const newState: State = produce(state, (draft) => {
+        draft.price += 5
+        draft.todos.push({
+            title: "hi",
+            done: true
+        })
+    })
+
+    const reducer = (draft: State) => {
+        draft.price += 5
+        draft.todos.push({
+            title: "hi",
+            done: true
+        })
+    }
+    const newState2: Immutable<State> = produce(reducer)(state)
+    const newState3: Immutable<State> = produce(reducer, state)()
+})
+
+it("can work with readonly base types", () => {
+    type State = {
+        readonly price: number;
+        readonly todos: readonly {
+            readonly title: string;
+            readonly done: boolean;
+        }[];
+    }
+
+    const state: State = {
+        price: 10,
+        todos: [{
+            title: "test",
+            done: false
+        }]
+    }
+
+    const newState: State = produce(state, (draft) => {
+        draft.price + 5
+        draft.todos.push({
+            title: "hi",
+            done: true
+        })
+    })
+
+    const reducer = (draft: Draft<State>) => {
+        draft.price += 5
+        draft.todos.push({
+            title: "hi",
+            done: true
+        })
+    }
+    const newState2: State = produce(reducer)(state)
+    const newState3: State = produce(reducer, state)()
 })
--- a/__tests__/produce.ts
+++ b/__tests__/produce.ts
@@ -74,16 +74,26 @@ it("can infer state type from default state", () => {
     type Recipe = (base?: State | boolean) => State
 
     let foo = produce((x: any) => {}, {} as State)
-    exactType(foo, {} as Recipe)
+    
+    type Arg0 = Parameters<typeof foo>[0]
+    type L = Parameters<typeof foo>["length"]
+    type Return = ReturnType<typeof foo>
+    exactType({} as Arg0, {} as (State | undefined))
+    exactType({} as any as L, 0 as (0 | 1))
+    exactType({} as Return, {} as State)
 })
 
 it("can infer state type from recipe function", () => {
     type State = {readonly a: string} | {readonly b: string}
-    type Recipe = (base: State, n: number) => State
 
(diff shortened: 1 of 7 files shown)


src/immer.d.ts
```diff
@@ -12,27 +12,17 @@ type AtomicObject =
...
+export type Draft<T> = T extends AtomicObject
     ? T
-    : T extends ReadonlyArray<any>
-    ? T[number][] extends T
-        ? DraftArray<T>
-        : DraftTuple<T>
-    : T extends AtomicObject
+    : T extends object
+    ? {-readonly [K in keyof T]: Draft<T[K]>}
+    : T // mostly: unknown & any
```
aleclarson: And also primitives like `number | string | boolean` 😛 Or am I misunderstanding this comment?

__tests__/produce.ts
```diff
@@ -70,27 +71,47 @@ it("can update readonly state via standard api", () => {
 // NOTE: only when the function type is inferred
 it("can infer state type from default state", () => {
     type State = {readonly a:number} | boolean
-    type Recipe = (base?: State | undefined) => State
+    type Recipe = (base?: State | boolean) => State
```
mweststrate: doh, that should just be `State`

__tests__/produce.ts
```diff
@@ -70,27 +71,47 @@ it("can update readonly state via standard api", () => {
...
     exactType(foo, {} as Recipe)
 })
 
 it("cannot infer state type when the function type and default state are missing", () => {
-    const res = produce(_ => {})
+    const res = produce((_: any) => {})
     exactType(res, {} as (base: any) => any)
+
+    // slightly different type inference...
+    const res2 = produce((_) => {})
```
mweststrate: this case is a bit weird, but `(_: any) => ` is seen by TS as a single arg function, while `_ =>` is variadic...

aleclarson:
I'll give it a thorough look sometime soon. Is there a reason you didn't just commit to #325?

Did you try out the curried producer changes I made in #325? If so, what didn't work about them?

src/immer.d.ts
```diff
@@ -92,32 +67,32 @@ export interface IProduce {
...
+        Recipe extends (...args: any[]) => any,
+        Params extends any[] = Parameters<Recipe>,
+        T = Params[0],
+    >(
+        recipe: Recipe,
+        initialState: T
+    ): (state?: Immutable<T>, ...rest: Tail<Params>) => Produced<Immutable<T>, ReturnType<Recipe>>
+
+    /** Normal producer */
+    <Base, D = Draft<Base>, Return = void>(
```
mweststrate: Separating Base and D allows things like: `produce<readonly string[]>([], (draft: string[]) => ...`

src/immer.d.ts
mweststrate: eh yes, had a `never` first there, and `AtomicObject` at the start, will remove the comment

mweststrate:
yes, this was based on that branch. Main problem there was that the rest arguments became optional in the returned producer function. Beyond that not too much changed, but I think the inference flow is slightly clearer this way, in which case we can better document certain best practices.

src/immer.d.ts
```diff
@@ -92,32 +67,32 @@ export interface IProduce {
...
-    ): Rest[number][] extends Rest | never[]
-        ? (
-              // The `base` argument is optional when `Rest` is optional.
-              base?: Immutable<Base>,
-              ...rest: Rest
-          ) => Produced<Immutable<Base>, Return>
-        : (
-              // The `base` argument is required when `Rest` is required.
-              base: Immutable<Base> | undefined,
-              ...rest: Rest
```
mweststrate: in practice, this `...rest: Rest` always resolved into `...any[]`, as soon as `Base` is set (not inferred) by using `produce<X>((draft, x: number) => {})`

__tests__/produce.ts
```diff
@@ -274,3 +297,68 @@ it("works with generic parameters", () => {
     let arr: ReadonlyArray<typeof val> = 0 as any
     insert(arr, 0, val)
 })
+
```
mweststrate: In these two test cases I tried to capture two real-life scenarios:
1. Your generic state shape is defined as mutable
2. The generic state shape is defined as immutable

Given those two scenarios (tests), the tests below verify a few things:
1. draft is always mutable, regardless the base state
2. if `state: State`, then ideally `produce(state, fn)` and `produce(fn)(state)` should result in `State` again

mweststrate: Ok, I think in the latest push I've found the solution that satisfies all those constraints! In the latest, the curried producer has it's own generic argument, and basically, what goes in, goes out.  That means: immutable types in, immutable types out. Mutable types in, mutable types out. However, the generic argument has to at least satisfy that it is at least the `Immutable<T>` where `T` is the first argument of the recipe. 

I think this solves all cases quite neatly. The only downside is that the tests that use `exactType` on the curried function now fails, because the curried producer has a wider type than the recipe. But technically that is actually valid, as the following statement is correct:

`produce((draft: { count: 0 }) => { draft.count++ }) ({ count: 7, messages: [] })`

The above now neatly does produce a new state of type `{ count, messages }` rather than just `{ count } ` which it would be, based on the recipe

mweststrate: cc @aleclarson

mweststrate:
@aleclarson I think this PR is done. I think this should be released as minor, although typescript types change (and TS 3.4 is probably required), there are no semantic or api changes to the library itself.

mweststrate:
...or I just finish 3.0 and merge it in there :)

jineshshah36:
Just my two cents:

You could use the types versions system in package.json to make this a non-breaking change. (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions)

Otherwise, this would break a lot of people’s code.

aleclarson:
@jineshshah36 I mentioned that [here](https://github.com/immerjs/immer/pull/325#issuecomment-478258783). If we do that, we need to use `git apply` in the CI build in order to test both versions.

aleclarson:
@mweststrate I'll see if I can review this today!

jineshshah36:
Would it be too much to keep two separate test files within versions in them like the types would be, and then, you only run the file with the matching version?

aleclarson:
@jineshshah36 That would require more maintenance than going with `git apply` and maintaining only the differences between 3.4 and 3.3. Honestly, I like the idea of supporting only TypeScript 3.4+ in Immer 3, like Michel mentioned.

mweststrate:
Closing this branch as it it is merged to `immer3` branch :)

PR #350, opened 2019-04-17
Feature Added: es6 maps support

keenondrums:


keenondrums:
It covers only ES6 Maps. I'm going to work on ES6 Sets, but Sets will have certain design challenges, so I would really really like to leave them for a separate PR.

__tests__/patch.js
```diff
@@ -118,6 +150,30 @@ describe("delete 1", () => {
     )
 })
 
+describe("delete 2", () => {
+    runPatchTest(
+        new Map([["x", 1]]),
+        d => {
+            d.delete("x")
+        },
+        [{op: "remove", path: ["x"]}],
```
aleclarson: Are `Map` entries assumed to be JSON-compatible?

src/common.js
```diff
@@ -30,16 +31,36 @@ export function original(value) {
...
+        for (let key in override) {
+            if (has(override, key)) {
+                target[key] = override[key]
+            }
+        }
+    })
+    return target
+}
+export function assign(target) {
+    const overrides = Array.prototype.slice.call(arguments, 1)
```
aleclarson: You can do `assign(target, ...overrides)` here

__tests__/patch.js
keenondrums: I guess one could serialize a Map to an object if needed. So not, not out-of-the-box. but yes with certain additional work.

src/common.js
```diff
@@ -30,16 +31,36 @@ export function original(value) {
...
+    if (isMap(target)) {
+        return assignMap(target, overrides)
     }
+    if (Object.assign) {
+        return Object.assign.apply(this, Array.prototype.slice.call(arguments))
```
aleclarson: This should be `Object.assign(target, ...overrides)`

src/common.js
```diff
@@ -80,11 +102,11 @@ export function shallowCopy(base, invokeGetters = false) {
 }
 
 export function each(value, cb) {
-    if (Array.isArray(value)) {
-        for (let i = 0; i < value.length; i++) cb(i, value[i], value)
-    } else {
-        ownKeys(value).forEach(key => cb(key, value[key], value))
+    if (Array.isArray(value) || isMap(value)) {
+        value.forEach((valueI, index) => cb(index, valueI, value))
```
aleclarson: nit: Maybe `entry` instead of `valueI`?

src/common.js
keenondrums: Wouldn't it fail for ES5 environments?

src/patches.js
```diff
@@ -1,9 +1,28 @@
-import {each} from "./common"
+import {each, isMap} from "./common"
 
 export function generatePatches(state, basePath, patches, inversePatches) {
-    Array.isArray(state.base)
-        ? generateArrayPatches(state, basePath, patches, inversePatches)
-        : generateObjectPatches(state, basePath, patches, inversePatches)
+    let generatePatchesFn = generatePatchesFromAssigned(
```
aleclarson: The flow should be:
```js
generatePatchesFn =
  isArray ? generateArrayPatches
  : isMap ? ...
  // isObject
  : ...
```
And please hoist the functions passed to `generatePatchesFromAssigned` so they can be reused. 👍

src/patches.js
```diff
@@ -1,9 +1,28 @@
...
+            return key in obj
+        }
+    )
+    if (Array.isArray(state.base)) {
+        generatePatchesFn = generateArrayPatches
+    }
+    if (isMap(state.base)) {
+        generatePatchesFn = generatePatchesFromAssigned(
+            function(map, key) {
+                return map.get(key)
```
aleclarson: These functions would look cleaner with arrow functions

src/patches.js
```diff
@@ -103,29 +128,55 @@ export function applyPatches(draft, patches) {
...
+                    return
+                }
+                base[key] = value
+            }
+            function add(key, value) {
+                if (isMap(base.base)) {
+                    base.set(key, value)
+                    return
+                }
+                if (Array.isArray(base)) {
```
aleclarson: nit: Probably want `isArray` check before `isMap` since it's more likely to be true

src/patches.js
```diff
@@ -103,29 +128,55 @@ export function applyPatches(draft, patches) {
...
+                    return
+                }
+                base[key] = value
+            }
+            function remove(key) {
+                if (isMap(base.base)) {
+                    base.delete(key)
+                    return
+                }
+                if (Array.isArray(base)) {
```
aleclarson: Same here

src/proxy.js
```diff
@@ -39,11 +40,17 @@ export function createProxy(base, parent) {
...
-          // although state itself is an object
-          Proxy.revocable([state], arrayTraps)
-        : Proxy.revocable(state, objectTraps)
+    let proxyTarget = state
+    let traps = objectTraps
+    if (Array.isArray(base)) {
+        proxyTarget = [state]
+        traps = arrayTraps
+    }
+    if (isMap(base)) {
```
aleclarson: Should be `else if`

src/common.js
aleclarson: Immer is compiled to ES5 by [bili](https://github.com/egoist/bili), so no. ;)

keenondrums:
@aleclarson addressed your comments. Thanks for the prompt review!
Currently, the build is failing because of the generator function. Working on it.

keenondrums:
@aleclarson I made Buble pass generators through, so you could see all the tests passing. Though it's still not a solution. Makes me think if nobody could think of a better way to wrap iterators for Maps, `immer` would have to go back to doing separate build for ES5 envs. Or not to support Maps.
What do you think?

aleclarson:
I'm working on changes that will include a fix to the generator issue. Then we can publish this!

keenondrums:
@aleclarson are thinking of implementing iterator protocol yourself?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

aleclarson:
Here's my work so far: https://github.com/immerjs/immer/commits/v3.1

There is a strange recursion bug that I'm trying to track down. Could you take a look?

keenondrums:
Will be pleased to do it tomorrow. It's 1AM where I am :)

keenondrums:
Closing it in favor of https://github.com/immerjs/immer/pull/351

PR #351, opened 2019-04-18
fix: reset bili.config, add Map getter for DRAFT_STATE, make isEnumerable more resilient

keenondrums:


diazig:
copy.clear()
    },
    forEach: (state, _, receiver) => (cb, thisArg) =>
        source(state).forEach((_, key, map) => {
            const value = receiver.get(key)
            cb.call(thisArg, value, key, map)
        }),
    get: state => key => {
        if (!state.modified && has(state.drafts, key)) {
            return state.drafts[key]
        }
        if (state.modified && state.copy.has(key)) {
            return state.copy.get(key)
        }
        const value = source(state).get(key)
        if (state.finalized || !isDraftable(value)) {
            return value
        }
        const draft = createProxy(value, state)

@keenondrums pushed 1 commit.

src/common.js
```diff
@@ -109,7 +109,8 @@ export function each(value, cb) {
 }
 
 export function isEnumerable(base, prop) {
-    return Object.getOwnPropertyDescriptor(base, prop).enumerable
+    const descriptor = Object.getOwnPropertyDescriptor(base, prop)
+    return descriptor && descriptor.enumerable
```
aleclarson: Was this causing a crash? I don't think we ever pass a property that doesn't exist. Or is this just for posterity?

src/proxy.js
```diff
@@ -185,6 +185,7 @@ const mapTraps = {
     }
 }
 const mapGetters = {
+    [DRAFT_STATE]: state => state,
```
aleclarson: Ah, thanks for catching this!

src/common.js
keenondrums: The descriptor is going to be undefined for Maps here https://github.com/immerjs/immer/blob/v3.1/src/immer.js#L288

src/proxy.js
keenondrums: Np. Thanks for refactoring the whole thing!

PR #352, opened 2019-04-18
v3.1.0

aleclarson:
Thanks to @keenondrums for adding `Map` support. ❤️


mweststrate:
I don't think the current map implementation is correct, I tried running some tests from #149 on current master, and they all fail. I'll try to push a PR with more tests shortly

mweststrate:
See #353

PR #353, opened 2019-04-18
Maps are broken

mweststrate:
Just as a quick check, I took some of the old tests of the earlier closest PR to add map support (#149), and they all seem to fail, although the look superficially speaking correct to me. Am I missing something obvious?

cc @aleclarson @keenondrums

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -1359,6 +1359,129 @@ function runBaseTest(name, useProxies, autoFreeze, useListener) {
             expect(isDraft(array)).toBeFalsy()
         })
     })
+
+    describe(`maps - ${name}`, () => {
+        it("should support Maps as root state", () => {
+            const base = new Map([["type1", "Finnish"]])
+            const array = [1, 2, 3, 4]
+            const result = produce(base, draft => {
+                draft.delete("type1")
+                draft.set("type2", "Italian").set("array", array)
+                return draft
+            })
+            expect(result.get("type1")).toBe(undefined)
+            expect(result.get("type2")).toBe("Italian")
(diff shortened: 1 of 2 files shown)


aleclarson:
The tests should be wrapped with `if (useProxies)`

Once that was added, only two of these tests fail

mweststrate:
ES5 is not supported? Then we should fix that first before releasing this
to the public. Shall we unpublish 3.1.0 for now?

aleclarson:
@mweststrate I see ES5 support as an enhancement more than a requirement. Users should still avoid `Map` objects if they need to target ES5. I can add a warning in the readme.

mweststrate:
No, that is a no-go area; Immer does support and maps can be polyfilled no
problem for IE 11. Proxies are a different ball game. The result of this
change is that devs can work happily for months on there immer + map
projects while devving. Having it blown into pieces the first time a IE
using customer arrives...

Maps should be supported in the ES5 implementation as well, or not be
released to avoid a lot of confusion and angry bug reports.

mweststrate:
Also note that adding map support is a breaking change, as having maps in a
draft is going to behave entirely different now :). Unpublishing and
reverting the MR for now before someone trips over it.

aleclarson:
I made a whoopsie! :P

aleclarson:
Also forgot to update the TypeScript defs :S

mweststrate:
Could you revert and open a new PR? Let's take such a fundamental change a
bit slower next time :)

aleclarson:
https://github.com/immerjs/immer/pull/354

Issue #146, opened 2018-04-27
Support for Set and Map


cmmartin:
* [x] Feature request: Is it possible to support mutating the native `Set`, `Map`, `WeakMap`, etc?
  * [ ] I am willing to implement this myself

RikuVan:
I had a quick try at adding Set and Map for es6 environment. I don't completely get everything happening in immer, so I well may have missed some complexities. It at first glance it seems actually pretty easy to proxy these since all mutations happen through method calls: 
```js
const nonMutatingMethods = [
    "forEach",
    "get",
    "has",
    "keys",
    "set",
    "values",
    "entries",
    "@@iterator"
]

const mutatingMethods = ["set", "delete", "clear", "add"]

const properties = ["size"]

function markMapOrSetChanged(proxy) {
    proxy.modified = true
    proxy.copy = shallowCopy(proxy.base)  // e.g. new Map(proxy.base)
    if (proxy.parent) markChanged(proxy.parent)
    nonMutatingMethods.forEach(function(method) {
        proxy[method] = function(a, b) {
            return proxy.copy[method](a, b)
        }
    })
    properties.forEach(function(prop) {
        proxy[prop] = proxy.copy[prop]
    })
    return proxy
}

function proxyMapOrSet(state) {
    var proxyState = {}
    var proxy = state
    nonMutatingMethods.forEach(function(method) {
        proxy[method] = function(a, b) {
            return proxy.base[method](a, b)
        }
    })
    mutatingMethods.forEach(function(method) {
        proxy[method] = function(key, value) {
            if (!proxy.modified) markMapOrSetChanged(proxy)
            proxy.copy[method](key, value)
        }
    })
    Object.defineProperty(proxy, PROXY_STATE, {
        value: proxy,
        enumerable: false,
        writable: true
    })
    proxyState.proxy = proxy
    return proxyState
}
```
I guess you don't need to recurse into the Map/Set values since they are all set through method calls?

RikuVan:
but it seems you can't freeze these, which means you have to do something kind of hacky there I guess:
```
function freezeMapOrSet(value) {
    const insertMethod = isMap(value) ? "set" : "add"
    if (autoFreeze) {
        ;[insertMethod, "delete", "clear"].forEach(function(method) {
            Object.defineProperty(value, method, {
                value: function() {
                    throw "In a frozen state, cannot be mutated"
                },
                enumerable: false,
                writable: false
            })
        })
        Object.freeze(value)
    }
    return value
[...]

PR #354, opened 2019-04-18
Map & Set support (v5)

aleclarson:
# Try v5.0.0 today!

```
npm install immer@next
```

### Tasks

- [x] ES6 `Map` / `Set` support with `setUseProxies(true)`
- [x] Ensure `onDelete`/ `onAssign` hooks work with ES6 maps/sets (and write tests)
- [x] Ensure object keys can be used with ES6 maps
- [x] ES6 `Map` / `Set` support in ES5 mode
- [x] Update typings for ES6 maps/sets
- [x] Update docs website
- [x] fix build
- [x] process review comments, create follow up tasks, also for typings
- [x] verify build size: `4.59 -> 6.28` gzipped, minified
- [x] check for duplicate code in traps, source fn
- [x] check for unnecessary closures in traps etc
- [x] how to freeze map / set?
- [x] check perf impact -> no significant changes


mweststrate:
Might be nice to have the additional tests from #353 in this branch, in case we missed something?

keenondrums:
@mweststrate I wish you had told me you wouldn't let the PR for ES6 only maps to pass through in [that thread](https://github.com/immerjs/immer/issues/146#issuecomment-483044865). Though, yeah, it all happened really fast :)
P.S. I have a working ES6 POC for Sets. Will raise a PR with it for this branch soon.

keenondrums:
Do we want to have ES5 support or do we want to have IE11 support? I believe it's not the same thing. IE11 doesn't support iterators and lots of other methods according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility).
This would still require to run some tests conditionally. It still would be an issue if a project, successfully using Maps, all of a sudden has to start supporting IE11. That kind of project still would have to be refactored.
@mweststrate are you firm on not letting ES6 only implementation pass through? Since it's gonna be a major release anyway, I guess it's ok to break things. Though IE11 folks wouldn't even be broken, they'll just have to stick to the old way of working with Maps manually. What we could do in this case is to add tests that the old way still works as expected.

mweststrate:
quick update: I'll hope to get back to it next week, travelling this week,
so might not have time to look into it / answer the questions :)

aleclarson:
Added patch-related functionality for ES6 Sets.

Items to discuss:
1. Patches for Maps will work only for string and numeric keys. Is it ok? Should we add a section to README only or have an exception thrown?
1. I'm not sure how compatible with [RFC-6902 JSON patch standard](http://tools.ietf.org/html/rfc6902) is [this record for ES6 Set patches](https://github.com/immerjs/immer/pull/358/files#diff-1bf9c25910eac0d559b090e8dae0dc0aR591). Currently, these patches cannot be applied in a different order.
1. Do we really-really-really want to support ES5? :)

TODOs:
- Typings
- Add tests (and most probably add missing implementation) for `onDelete` and `onAssign` hooks for cases with ES6 Maps and ES6 Sets

@aleclarson if you merge it to v4 branch, I'd be happy to carry on this discussion there, since it really does belong there.

_Originally posted by @keenondrums in https://github.com/immerjs/immer/pull/358#issuecomment-485542670_

aleclarson:
> 1. Patches for Maps will work only for string and numeric keys. Is it ok? Should we add a section to README only or have an exception thrown?

No, we should make sure changes to object keys have patches created for them, since Immer patches can be useful even when not serialized to JSON. 

__Note:__ Immer prefers working with tree-like data structures, and I'm not sure what the bugs are when relational data is used.

> 2\. I'm not sure how compatible with [RFC-6902 JSON patch standard](http://tools.ietf.org/html/rfc6902) is [this record for ES6 Set patches](https://github.com/immerjs/immer/pull/358/files#diff-1bf9c25910eac0d559b090e8dae0dc0aR591). Currently, these patches cannot be applied in a different order.

To be clear, Immer patches aren't strictly JSON patches, but you must use JSON-only keys/values if you ever need to serialize the patches.

> 3\. Do we really-really-really want to support ES5? :)

To not support ES5 would add unwanted complexity to the API for ES5 users. :(

_Originally posted by @aleclarson in https://github.com/immerjs/immer/pull/358#issuecomment-485575527_

keenondrums:
@aleclarson do we want to support ES5 or IE11? It's not the same thing. As [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) states IE11 doesn't support `new Map([iterable])` which means we'll have to come up with a different mechanism to shallow copy Maps in IE11. 
Moreover, we do not run tests specifically for IE 11 using that very IE11. It means we can not really guarantee we support it.

mweststrate:
Some random thoughts:

I think it is fine to limit maps to only allow strings / numbers as keys. I think having objects as keys, which can then be updated in a producer as well, will get confusing very quickly, and just bailing out is probably better until we encounter a good RL use case

The goal is to make sure devs can use immer safely inside consumer facing applications, without sudden suprises. In practice that means that we need to support ES5, or at least, things that can be polyfilled to ES 5. So in this specific case, whether `Map([iterable])` isn't that important, the important question is, does the implementation correctly after being processed by babel? If babel polyfills the map constructor correctly, it is fine. In MobX there are two major versions still being maintained, which is also not ideal. But at least, practically all new functionality in 5 can be expressed in mobx 4, without big semantic changes. 

The risky semantic difference here is that sets and maps would be treated totally different as soon as they are considered draftable. So it could be an option to only support sets and maps in a next major, but I think it would be a significant maintenance burden. Mostly being helping people out that, ignorantly, npm installed the latest version, work a while with that, use maps / sets (probably without checking the docs on that at all seemed to works ootb), and then discover that they have to rewrite a lot of code after a first round of QA on IE11...

omninonsense:
> 1. Patches for Maps will work only for string and numeric keys. Is it ok? Should we add a section to README only or have an exception thrown?

I don't think there's much valid use-cases outside of strings/numbers anyway. I think the main benefit of using a `Map` over an object is convenience (ie numbers not being coerced into strings, easier/safer preservation of type information in Flow/TS).

shabbyrobe:
> I don't think there's much valid use-cases outside of strings/numbers anyway.

It's actually a _very_ important use-case for Sets and Maps. Most of the time I reach for them it's for this property.

keenondrums:
@shabbyrobe I have to correct my statement. I think any valid key (including objects, symbols and etc.) is going to work with Maps. I just doubt if it's going to work with patches enabled.

omninonsense:
> It's actually a _very_ important use-case for Sets and Maps. Most of the time I reach for them it's for this property.

Sorry for the confusion. To give it a bit of context: I made my comment with redux in mind. I think a best practice is to avoid putting non-serializable data inside the state.

shabbyrobe:
> I think any valid key (including objects, symbols and etc.) is going to work with Maps.

Ah that's good news! Thank you for the clarification.

> I made my comment with redux in mind. I think a best practice is to avoid putting non-serializable data inside the state.

That makes complete sense from a Redux point of view. Immer doesn't force you to use Redux though does it? I had been thinking of introducing it to a non-React/Redux app soon as I think it can help with a few data management issues there too.

> I just doubt if it's going to work with patches enabled.

Is this because they're JSON Patch patches? If so, that makes sense and seems like a totally reasonable constraint to me (others may feel differently of course!)

aleclarson:
> I think any valid key (including objects, symbols and etc.) is going to work with Maps. I just doubt if it's going to work with patches enabled.

There's nothing stopping you from using non-serializable data in Immer patches. What makes you think otherwise?

keenondrums:
Maybe, you're right. I'll try to add tests to cover that over the weekend.

mweststrate:
I think indeed that any key type should work (especially since the patch
paths in immer are not plain strings, but an array of path segments), but
that we should warn that using arbitrary keys might result in
not-by-default-serializable patches.

A second consideration, for which I cannot imagine a real actual use case,
is that key objects themselves are not being proxied IIRC? I don't think
that is a problem in practice, but it means that something like `for (const
key of map.keys()) key.count++` might not behave properly

aleclarson:
> for (const key of map.keys()) key.count++

@keenondrums Can you add a test for this while you're at it? :D

keenondrums:
@aleclarson I'm pretty sure it's going to fail. Keys of maps are not proxied. Though I don't see it as an issue.

keenondrums:
@aleclarson I'm currently working on hooks for Maps and Sets and I found out that hooks never get a new top-level object. Is it expected?

```js
produce({a: 0, b: 0, c: 0}, s => {
  s.a++
})
```
Calls onAssign with `["a",1]`. Shouldn't it be `["a", 1]` and` [undefined, {a: 1, b: 0, c: 0}]`?

aleclarson:
Yeah, that's expected. Is there a reason you would want it that way? Currently, the `onCopy` hook is called with `[{a: 1, b: 0, c: 0}]`.

keenondrums:
I was just wondering. Thanks!

Do you want to pass changes of nested values inside of Sets to `onAssign`? I, personally, see little or no value of doing that but I have never used hooks (it can be done from a technical perspective though if necessary).

In the words of code:
```js
produce({a: new Set([1])}, s => {
  s.a.add(2)
})
```
It would call onAssign one time only with `['a', Set { 1, 2 }]`.

My concern is that adding `onAssign` calls for changes of data inside of Sets could be confusing because Sets don't have keys.
In other words, if we added that functionality, `onAssign` would be called two times with: `[2, 2]` and  `['a', Set { 1, 2 }]`.

keenondrums:
@aleclarson [here's a PR with non-primitive keys for Maps and hooks for Maps and Sets](https://github.com/immerjs/immer/pull/370). Take a close look at tests for hooks to make sure we're on the same page regarding `onDelete` and `onAssgin` for `Sets`.

aleclarson:
@keenondrums Good stuff, thanks for your hard work! 👍

aleclarson:
# Try v4.0.0 today!

```
npm install immer@next
```

mweststrate:
What is the current status of this PR?

aleclarson:
@mweststrate Limbo until someone implements ES5 support for `Map` and `Set`.

keenondrums:
Guys, I'm terribly sorry. I've been kind of busy lately. I should have some time to work on it next week.

mweststrate:
Thanks for the update, no worries! The work so far is very much
appreciated, so take your time :)

mweststrate:
Currently working on the merge conflicts, might take a while. Seems a a prettier config change was somewhere accidentally accepted while this branch was alive.😞

aleclarson:
@mweststrate Ah that must've been when I set `"useTabs": true`. Sorry about that!

mweststrate:
I resolved the merge conflicts. There might be some mistakes in there, as it was quite tricky. Some tests related to `patches` are still failing at least, most notable is this error (does it ring a bell @runnez)?

```
TypeError: Method Map.prototype.forEach called on incompatible receiver #<Map>
        at Map.forEach (<anonymous>)

      113 | export function each(obj, iter) {
      114 | 	if (Array.isArray(obj) || isMap(obj) || isSet(obj)) {
    > 115 | 		obj.forEach((entry, index) => iter(index, entry, obj))
          | 		    ^
      116 | 	} else {
      117 | 		ownKeys(obj).forEach(key => iter(key, obj[key], obj))
      118 | 	}

      at forEach (src/common.js:115:7)
```

Coming days I won't have time to further investigate, so I'll continue after that, or if anyone beats me to it, welcome :). 

Sorry, this got a lot more complicated than it ought to be as there were significant changes on master as well, but it is tricky to find them one by one due to the whitespace changes. 

I think it should be possible to build the traps up a bit more efficiently, but let's get the tests passing first.

aleclarson:
@mweststrate Git has options for merging without whitespace conflicts (see [here](https://stackoverflow.com/questions/9776527/merging-without-whitespace-conflicts)). Maybe you should rebase and keep the tabs in place?

mweststrate:
Good one, gonna try that as well.

runnez:
@mweststrate I've found a place where logic is broken. If we move back an old function, it will fail but only one test. I'll try to fix it tomorrow 
https://github.com/immerjs/immer/blob/ae80361f51e07fa05c4c73f8fa3554361a202732/src/patches.js#L141
```
  ● applyPatches › applied patches cannot be modified

    TypeError: Cannot add/remove sealed array elements
        at Array.splice (<anonymous>)

      154 | 			const add = (key, value) =>
      155 | 				Array.isArray(base)
    > 156 | 					? base.splice(key, 0, value)
          | 					       ^
      157 | 					: isMap(base)
      158 | 					? base.set(key, value)
      159 | 					: isSet(base)

      at splice (src/patches.js:156:13)
      at add (src/patches.js:176:6)
      at src/immer.js:166:4
      at Immer.recipe (src/immer.js:74:14)
      at Immer.produce [as applyPatches] (src/immer.js:165:15)
      at Object.<anonymous> (__tests__/patch.js:105:3)

 PASS  __tests__/base.js

Test Suites: 1 failed, 10 passed, 11 total
Tests:       1 failed, 1730 passed, 1731 total
Snapshots:   143 passed, 143 total
Time:        4.014s
```

runnez:
@mweststrate I've fixed tests, but I've got doubts about expected behaviour on this line https://github.com/immerjs/immer/pull/354/commits/ff25a83820815fd5672a33d4c033f40f3805ed58#diff-b40ab5eb20dfada012e789db3e9780bbR142

runnez:
@mweststrate I've fixed errors in ie11, but you may revert it if need https://github.com/immerjs/immer/pull/354/commits/ffd8769f51f1769d3fd08b08a35ad14ffe37f95a

src/proxy.js
```diff
@@ -258,7 +259,7 @@ const mapTraps = makeTrapsForGetters({
 	keys: state => () => latest(state).keys(),
 	values: iterateMapValues,
 	entries: iterateMapValues,
-	[Symbol.iterator]: iterateMapValues
+	[hasSymbol && Symbol.iterator]: iterateMapValues
```
mweststrate: I think `hasSymbol ? Symbol.iterator : "@@iterator"` would be better; it avoids creating "null" keys, and I've seen in other libs that `@@standardSymbolName` is a common fallback for the official symbols

src/proxy.js
```diff
@@ -296,7 +297,7 @@ const setTraps = makeTrapsForGetters({
 	keys: iterateSetValues,
 	values: iterateSetValues,
 	entries: iterateSetValues,
-	[Symbol.iterator]: iterateSetValues
+	[hasSymbol && Symbol.iterator]: iterateSetValues
```
mweststrate: Same

mweststrate:
@runnez thanks for investigating! Sorry for messing up that if-else logic while merging 😅

src/proxy.js
runnez: I agree, it looks cleaner. But if env doesn't have Symbol, these code won't fire. Or Did you mean only readability?

mweststrate:
No I mean it is actually safe and a good idea to set is to a property
called "@@iterator" as fallback, as there are a couple of utility libraries
that actually look for that in the absence of a proper Symbol.iterator

runnez:
> I think the condition is incorrect; even if the base is a set, and the op is add, the patch should be defended against accidental mutation, as theoretically a next patch could change that value _inside_ that set (and otherwise, if we are going to error anyway, it doesn't matter that we didn't optimize this clone anymore)

@mweststrate did I understand you correctly? https://github.com/immerjs/immer/pull/354/commits/24658bc36e0d0bb09d8df5e2a08ecc51fde916d3

__tests__/patch.js
```diff
@@ -388,6 +581,72 @@ describe("arrays - splice (shrink)", () => {
...
+	runPatchTest(
+		new Set([{id: 1, val: "We"}, {id: 2, val: "will"}]),
+		d => {
+			const obj1 = findById(d, 1)
+			const obj2 = findById(d, 2)
+			obj1.val = "rock"
+			obj2.val = "you"
+		},
+		[
+			{op: "remove", path: [0], value: {id: 1, val: "We"}},
```
mweststrate: I'd expect patches here that capture that the modification was only the object inside the sets, rather than patches on the set itself, like:
`op: "replace", path: [1, "val"], value: "will"`

Is there a special reason the entire set entry gets reconstructed? (I agree that the concept of indices in a set can be challenged, and maybe your approach is more correct, but sets guarantee insertion order, so I think it should be ok rely on paths in sets?)

src/es5.js
```diff
@@ -144,11 +158,166 @@ function proxyProperty(draft, prop, enumerable) {
...
+			return fn(state)
+		}
+	}
+}
+
+function proxyMethod(trap, key) {
+	return {
+		get() {
+			return (...args) => {
+				const state = this[DRAFT_STATE]
```
mweststrate: Using `this` in an arrow function, is this doing what it is supposed to be doing? Should be `function (...args)` I presume

mweststrate: If this is correct, a comment would be warranted :)

src/proxy.js
```diff
@@ -94,93 +190,178 @@ arrayTraps.set = function(state, prop, value) {
...
+	"preventExtensions",
+	"isExtensible",
+	"getPrototypeOf"
+])
+
+/**
+ * Map drafts
+ */
+
+const mapTraps = makeTrapsForGetters({
```
mweststrate: Random thought: could we use the ES5 implementation, and not use Proxies for maps and sets at all? If that would simplify the implementation that would be cool. If "ES5" administration doesn't play nicely inside a proxy draft tree, than lets leave it as is. But proxies themself don't add much for maps and sets, since all object interactions are through functions anyway.

__tests__/patch.js
mweststrate: I think this quite an edge case anyway, so we could also open it as a fresh follow up issue, and see if anybody runs into the difference, before doing anything.

src/proxy.js
mweststrate: Could also be opened as follow up optimization issue, to reduce the package size

__tests__/patch.js
aigoncharov: Ambiguity of indices in Sets was the primary reason. As you said, you can iterate over them, however, you cannot get an item at index.
Love the idea of leaving this edge case out of the PR.

src/es5.js
runnez: @mweststrate it provides context for `this`. When we call `someMap.forEach()`, it call `get() {...}` which return function

runnez: Hmm, I've tried to use function, and it works. I have no idea why)

src/proxy.js
runnez: I think it's possible to use one codebase for Map and Set and use both Proxy and ES5 mechanics. Let's do it as a follow up issue

mweststrate:
Perfect, I'll try to wrap up remaining things and cut a release tonight :)

src/proxy.js
mweststrate: See #449

src/es5.js
mweststrate: Updated, just in case. Don't get why it worked though :-P.

__tests__/patch.js
mweststrate: See #450

PR #358, opened 2019-04-22
feat: Add ES6 Sets support

keenondrums:


keenondrums:
This one doesn't have patch-related functionality yet. Will work on it as soon as I have some free time.

src/proxy.js
```diff
@@ -297,3 +402,16 @@ function makeIterable(next) {
...
+
+function makeTrapsForGetters(getters) {
+    return {
+        get(state, prop, receiver) {
+            return getters.hasOwnProperty(prop)
+                ? getters[prop](state, prop, receiver)
+                : Reflect.get(state, prop, receiver)
+        },
+        ownKeys(state) {
+            return Reflect.ownKeys(source(state))
```
aleclarson: This trap can be removed

src/proxy.js
```diff
@@ -282,9 +377,19 @@ function peek(draft, prop) {
 }
 
 function markChanged(state) {
+    let makeCopyFn = assign
```
aleclarson: `makeCopyFn` --> `assignFn`

src/proxy.js
```diff
@@ -282,9 +377,19 @@ function peek(draft, prop) {
...
+    } else if (isSet(state.base)) {
+        makeCopyFn = assignSet
+    }
     if (!state.modified) {
         state.modified = true
-        state.copy = assign(shallowCopy(state.base), state.drafts)
+        state.copy = makeCopyFn(shallowCopy(state.base), state.drafts)
+        // state.copies is present for Sets only
+        if (state.copies !== undefined) {
+            state.copies = state.drafts
```
aleclarson: If we can, we should use the `drafts` property after `modified = true` (for `Set` only) instead of adding a new `copies` property. Is there anything preventing us from doing that?

src/proxy.js
```diff
@@ -262,6 +272,91 @@ function iterateMapValues(state, prop, receiver) {
     }
 }
 
+/**
+ * Set drafts
+ */
+
+const setGetters = {
+    [DRAFT_STATE]: state => state,
+    size: state => source(state).size,
```
aleclarson: Looks like `mapGetters` and `setGetters` share some values.
```js
const setGetters = {
  ...mapGetters,
```
You can omit `[DRAFT_STATE]`, `size`, and `has` here.

src/proxy.js
```diff
@@ -262,6 +272,91 @@ function iterateMapValues(state, prop, receiver) {
...
+        return state.copy.delete(value)
+    },
+    clear: state => () => {
+        markChanged(state)
+        return state.copy.clear()
+    },
+    forEach: state => (cb, thisArg) => {
+        const iterator = iterateSetValues(state)()
+        let result = iterator.next()
+        while (!result.done) {
```
aleclarson: Might be able to use `mapGetters.forEach` too

src/proxy.js
```diff
@@ -262,6 +272,91 @@ function iterateMapValues(state, prop, receiver) {
...
+        }
+        return state.draft
+    },
+    delete: state => value => {
+        markChanged(state)
+        return state.copy.delete(value)
+    },
+    clear: state => () => {
+        markChanged(state)
+        return state.copy.clear()
```
aleclarson: Does this need to mutate `state.assigned`?

src/proxy.js
keenondrums: `drafts` is being set to `null`. I figured an additional property for Sets would be easier to understand than a conditional assignment of `drafts`  to null. Though I don't have any strict preference.

aleclarson: I think the existence of both `drafts` and `copies` is confusing, since `copies` may contain drafts.

src/proxy.js
keenondrums: For now it seems like we don't need `assigned` for ES6 Sets at all. It's used for other data structures to generate 'replace' patches, but we cannot really replace anything in a Set, we can only remove and add.

src/proxy.js
keenondrums: Wouldn't it return keys of our state then?

aleclarson: Good catch. 👍  I remember adding this trap to fix an internal issue, then removing it once realizing it's not necessary to fix the internals. I didn't think of users calling `Object.keys` on their Set/Map objects, which definitely shouldn't return the keys of the draft-state.

aleclarson: If we care about the `Object.keys` edge case, we should also support `Object.getOwnPropertyDescriptor` by adding a `getOwnPropertyDescriptor` trap.

aleclarson: And what about the `getPrototypeOf` or `setPrototypeOf` traps? And the `defineProperty`, `set`, `deleteProperty` traps? It gets to be a bit much, but for the sake of correctness, we should probably do it?

src/proxy.js
keenondrums: This would add other map getters (like `set`), which we don't have overrides for, to set getters (Set do not have `set` method). Anyway, it's only 3 of them and they are really small. If you're firm on this we certainly can figure out some way to share this code, but it's just a couple of lines, so I'm not sure it's worth it.

src/proxy.js
keenondrums: I guess we're gonna need tests to cover them as well.

src/proxy.js
aleclarson: Good point. We can leave it as-is

keenondrums:
@aleclarson addressed your comments: either made a corresponding change or replied in a thread.

keenondrums:
Added patch-related functionality for ES6 Sets.

Items to discuss:
1. Patches for Maps will work only for string and numeric keys. Is it ok? Should we add a section to README only or have an exception thrown?
1. I'm not sure how compatible with [RFC-6902 JSON patch standard](http://tools.ietf.org/html/rfc6902) [this](https://github.com/immerjs/immer/pull/358/files#diff-1bf9c25910eac0d559b090e8dae0dc0aR591) record for ES6 Set patches. Currently, these patches cannot be applied in a different order.
1. Do we really-really-really want to support ES5? :)

TODOs:
- Typings
- Add tests (and most probably add missing implementation) for `onDelete` and `onAssign` hooks for cases with ES6 Maps and ES6 Sets

aleclarson:
> 1. Patches for Maps will work only for string and numeric keys. Is it ok? Should we add a section to README only or have an exception thrown?

No, we should make sure changes to object keys have patches created for them, since Immer patches can be useful even when not serialized to JSON. 

__Note:__ Immer prefers working with tree-like data structures, and I'm not sure what the bugs are when relational data is used.

> 2\. I'm not sure how compatible with [RFC-6902 JSON patch standard](http://tools.ietf.org/html/rfc6902) is [this record for ES6 Set patches](https://github.com/immerjs/immer/pull/358/files#diff-1bf9c25910eac0d559b090e8dae0dc0aR591). Currently, these patches cannot be applied in a different order.

To be clear, Immer patches aren't strictly JSON patches, but you must use JSON-only keys/values if you ever need to serialize the patches.

> 3\. Do we really-really-really want to support ES5? :)

To not support ES5 would add unwanted complexity to the API for ES5 users. :(

aleclarson:
Merged into the `v4` branch

PR #366, opened 2019-05-14
Added check for Symbol.for for IE11 compatibility

faberyx:
Just added a  && System.for check to prevent immer to  brake on IE11


aleclarson:
Thanks ❤️

PR #384, opened 2019-06-07
Add unpkg and jsdelivr fields to package.json

tmcw:
Immer already has a UMD build, which is great! This PR adds jsdelivr and unpkg fields in package.json pointing to the same place, so jsdelivr.com and unpkg.com respectively can automatically serve the right files, and AMD implementations that depend on one or both of those (like d3-require) can automatically load immer directly or as a dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,8 @@
   "description": "Create your next immutable state by mutating the current one",
   "main": "dist/immer.js",
   "umd:main": "dist/immer.umd.js",
+  "unpkg": "dist/immer.umd.js",
+  "jsdelivr": "dist/immer.umd.js",
   "module": "dist/immer.module.js",
   "jsnext:main": "dist/immer.module.js",
   "react-native": "dist/immer.module.js",


aleclarson:
Hey thanks! I wonder why they don't support `umd:main`?

aleclarson:
Note: You can always use `https://bundle.run/immer` instead

tmcw:
There's a lot of bundle history so I might be missing something, but as far as I can tell, [microbundle](https://github.com/developit/microbundle/issues/91) coined umd:main, and unpkg predates it by quite a while, as does jsdelivr. Surprisingly nobody's asked about umd:main on the unpkg issue tracker, but also notably [microbundle switched to recommending the unpkg](https://github.com/developit/microbundle/pull/354) field. Like the rest of JavaScript modules, it's… chaotic.

Yep, though bundle.run works, it’s a bigger piece of infrastructure to maintain than the NPM CDNs, so it’s often slow or crashed.

aleclarson:
Cool, thanks for the insight!

aleclarson:
Maybe `jsdelivr` should be removed since Microbundle and Unpkg both support the `unpkg` field?

tmcw:
I think getting rid of `umd:main` should be fine because no consumers use it, but I think jsdelivr should stay because sadly they don't support the unpkg field - see [Tips for Module Authors](https://www.jsdelivr.com/features) on this page.

aleclarson:
@jimaek Yo, can you weigh in here?

aleclarson:
Oh, jsDelivr isn't open source, so I don't see why we should cater to them.

aleclarson:
How ubiquitous is jsDelivr?

aleclarson:
Meh, I've overblown this. I'm just gonna merge. 😂

PR #385, opened 2019-06-07
feat: export "produceRO" for 100% immutable returns

### LINKED ISSUE #381: Typescript with initialization without change
First of all, thanks for this great library, I already love immer :)
But one thing bothers me, I hope someone can help.

What I'm trying to do: initialize an object as immutable but **without** any real property changes. Not the object instance is important for me (so the fact that it's not 'freezed' doesn't matter, because I know, in production it's already turned off by default). The only thing is important for me: **to make Typescript catch any mistaken direct changes** to the properties after I 'made it immutable'.

But the strange thing is that: if I give an empty producer in calling 'produce' (CASE 1), after that Typescript allows to directly modify property of the returned 'immutable' object (what I don't want, and don't understand why)!
But...if I do any simple modification in the producer (CASE 2), Typescript will not allow direct changes on the properties of the returned object (as expected).

Why (and how) Typescript differentiates between the 'empty' and the 'really modifying' producer? How could I initialize the immutable object without any real changes but making Typescript after that to catch when trying to modify readonly properties?

Thanks in advance.

```typescript
import produce, { immerable } from "immer";

class TheClass {
  [immerable] = true;
  public title: string = "";
}

// Just initialize the object
let data = new TheClass();
// Want to make it IMMUTABLE from this point, to avoid any direct property changes

// CASE 1
let immutable1 = produce(data, (draft) => { /* empty, dont want to change anything */ });
// Typescript should NOT allow this, it should be immutable with readonly properties
immutable1.title = "shouldn't be able to modify..."; // But, it works....

// CASE 2
let immutable2 = produce(data, (draft) => { draft.title = "initial modification"; });
// Error (as expected): Cannot assign to read only property...
immutable2.title = "it won't work, Typescript catches it";
```

[codesandbox](https://codesandbox.io/s/empty-wildflower-l873t?autoresize=1&expanddevtools=1&fontsize=14&module=%2Fsrc%2Findex.ts)

aleclarson:
This provides a way to opt-in for `produce` to always wrap your base type with `Immutable<T>` so the return type is 100% immutable.

**Note:** TypeScript support for creating generic function types within conditionals is **not** possible yet.

Closes #381

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "rimraf": "^2.6.2",
     "seamless-immutable": "^7.1.3",
     "spec.ts": "^1.1.0",
-    "typescript": "3.4.3",
+    "typescript": "3.5.1",
     "yarn-or-npm": "^2.0.4"
   },
   "jest": {
--- a/src/immer.d.ts
+++ b/src/immer.d.ts
@@ -69,43 +69,60 @@ export type Produced<Base, Return> = Return extends void
  * @param {Function} patchListener - optional function that will be called with all the patches produced here
  * @returns {any} a new state, or the initial state if nothing was modified
  */
-export interface IProduce {
-    /** Curried producer */
-    <
-        Recipe extends (...args: any[]) => any,
-        Params extends any[] = Parameters<Recipe>,
-        T = Params[0]
-    >(
-        recipe: Recipe
-    ): <Base extends Immutable<T>>(
-        base: Base,
-        ...rest: Tail<Params>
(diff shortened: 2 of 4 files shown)


mweststrate:
5.3 will introduce an `castImmutable` utility, which can be combined with `produce` to produce an immutable type. That is not as straight forward as this proposal, but also keeps the api a bit smaller and more composable, so lets see if that suffices first. See also updated docs in the commit linked in #381

PR #395, opened 2019-06-29
auto-own createDraft drafts when assigning through proxies (#393)

### LINKED ISSUE #393: Use createDraft() inside produce() for better type-safety
## 🚀 Feature Proposal

The ability to use `createDraft` (or similar) inside `produce` to import immutable structures into a draft.

## Motivation

Currently there is no way to assign an immutable structure to a draft property. Further, if a `ReadonlyArray` is involved, type errors prevent this.

We're using this to copy an existing immutable structures into draft instances during a produce (e.g. from other object trees or initial values stored in global `const`s).

Minimal example:
```typescript
type Foo = Readonly<{
    a: number;
}>;
type Model = Readonly<{
    foo: Foo;
}>;

const immutableFoo: Foo = { a: 1 };
immutableFoo.a = 2; // type error

const model: Model = { foo: { a: 0 } };
const model2 = produce(model, (draft) => {
    draft.foo = immutableFoo; // no type error, but should!
    draft.foo.a = 3; // unexpectedly mutates immutableFoo! 😧
});
console.log(immutableFoo); // mutated!
console.log(model2);
```

With a `ReadonlyArray` we get expected type errors:
```typescript
type Foo = ReadonlyArray<number>;
type Model = Readonly<{
    foo: Foo;
}>;

const immutableFoo: Foo = [0];
immutableFoo.push(1); // type error

const model: Model = { foo: [] };
const model2 = produce(model, (draft) => {
    draft.foo = immutableFoo; // type error
    draft.foo.push(3); // unexpectedly mutates immutableFoo 😭
});
console.log(immutableFoo);
console.log(model2);
```

## Example 

It is almost solved by using the `createDraft` feature:
```typescript

type Foo = ReadonlyArray<number>;
type Model = Readonly<{
    foo: Foo;
}>;

const immutableFoo: Foo = [0];
const model: Model = { foo: [] };
const model2 = produce(model, (draft) => {
    draft.foo = createDraft(immutableFoo); // no more type error! 😀
    draft.foo.push(3);
});
console.log(immutableFoo); // not mutated! 😀😀
console.log(model2); // has the new value 😀😀😀
console.log(isDraft(model2.foo)); // still a draft 😞
```

I could either see `createDraft` "just working" or a new similar function (maybe `asDraft`) that provides this behavior (if the behavior doesn't make sense for `createDraft`).

benmosher:
Heads up, tests fail for the es5 implementation when assigning to new fields. I am not sure how to solve that. 

TBH, as I understand the es5 version, it's unsolvable. Feels like a potential compromise and/or footgun to just not support it for es5 because at the same time, it'd be great to have this working in environments that can use native proxies.

Fixes #393.

### DIFF
--- a/__tests__/manual.js
+++ b/__tests__/manual.js
@@ -132,5 +132,49 @@ function runTests(name, useProxies) {
 			finishDraft(draft)
 			expect(() => finishDraft(draft)).toThrowErrorMatchingSnapshot()
 		})
+
+		describe("attach to existing draft", () => {
+			describe("array assign", () => {
+				const b = [{foo: "bar"}, {foo: "baz"}]
+				const c = produce(b, bb => {
+					bb[1] = createDraft({foo: "quux"})
+					expect(isDraft(bb[1])).toBe(true)
+				})
+				expect(c.length).toEqual(2)
+				expect(isDraft(c[1])).toBe(false)
+			})
+			describe("array push", () => {
+				const b = [{foo: "bar"}, {foo: "baz"}]
+				const c = produce(b, bb => {
+					bb.push(createDraft({foo: "quux"}))
+					expect(isDraft(bb[2])).toBe(true)
+				})
+				expect(c.length).toEqual(3)
+				expect(isDraft(c[2])).toBe(false)
+			})
+			describe("object set existing", () => {
+				const b = {foo: {bar: "baz"}}
+				const c = produce(b, bb => {
+					const foo2 = createDraft(Object.freeze({bar: "quux"}))
+					expect(foo2.bar).toEqual("quux")
+
+					foo2.bar = "zip"
+					expect(foo2.bar).toEqual("zip")
+
+					bb.foo = foo2
+					expect(isDraft(bb.foo)).toBe(true)
+				})
+				expect(isDraft(c.foo)).toBe(false)
+				expect(c.foo.bar).toEqual("zip")
+			})
+			describe("object set new", () => {
+				const b = {foo: {bar: "baz"}}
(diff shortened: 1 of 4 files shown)


mweststrate:
I doubt this PR is a good idea. It confuses two different problems: casting and creating drafts. (And a it makes a static type checking problem a runtime problem, which is generally a really bad trade) Currently traveling so can't really elaborate, but please don't merge yet.

benmosher:
For the record, I'm not sure this exact solution is a good idea, either. It is just one of several possible solutions and I wanted to discuss over code and see what concrete problems shook out (like the es5 failures).

I thought it was just a static compilation problem at first, but it is also fixes a runtime problem: if you cast an object T to Draft<T>, and it has any readonly fields at any depth, you could try to mutate them later (in a forEach callback, for example) and the typing would check out but it would fail at runtime on the inserted objects. 

My first draft of this PR used a new asDraft function and only took ownership of drafts created by that function, but after thinking about it I figured the addition of a new public API cost more long term than this change to the behavior of createDraft. 

No problem if you don't want to merge this as is. I just figured having a solution to look at would help make the discussion more concrete. Also this project is great and I wanted to pitch in to say thanks! But as a maintainer myself I know every change incurs maintenance cost, and evaluating unsolicited PRs is nontrivial cognitive burden so no worries!

benmosher:
Edit: just caught up on discussion on #393. Looks like you are already talking through the stuff I mentioned. Seems like the simplest first step is to close this because the solution (if any) will not change the behavior of createDraft 👍🏼

PR #397, opened 2019-07-16
Include import statement in `original` example

snapwich:
All the other examples in the readme import the demonstrated immer export except `original`.  Added import for `original` example.


mweststrate:
Thanks!

Issue #400, opened 2019-07-31
API to simplify a producer + patches


mweststrate:
## 🚀 Feature Proposal

Introduce API's to capture the next state and patches directly:

* `produceWithPatches(baseState, recipe) => [nextState, patches, inversePatches]`
* `produceWithPatches(recipe) => (baseState, ...args) => [nextState, patches, inversePaches]` (curried form)

## Motivation

Currently it is quite annoying to grab the patches from a producer, for two reasons:

First, directly returning is often impossible, and a lot of temporarily local vars enter the game, in order to be able to wire up the receiving both the next state and the patches. The following pattern I encounter quite reqularly:

```javascript
const myCoolFunc(state) => produce(state, draft => { /* something */ })

// to capture patches, it must be destructred to :
// (this is even worse if you used a curried produce, as it has to be uncurried to be able to write the following)
const myCoolFuncWithPatches(state) => {
   let patches
   const nextState = produce(state, draft => { /* something */ }, p => { patches = p })
   return [nextState, patches]
}

// with this proposal it becomes
const myCoolFunc(state) => produceWithPatches(state, draft => { /* something */ })
```

cc @aleclarson

PR #402, opened 2019-08-02
feat: introduced produceWithPatches. Implements #400

mweststrate:
#400


mweststrate:
Alternative implementation could have been `produce(state, recipe, true)`, however, that doesn't provide a simple way to create a curried producer that captures patches (which is currently not possible)

PR #415, opened 2019-09-06
feat: Always use remove command in generateArrayPatches()

### LINKED ISSUE #208: Patch for array.pop() replaces "length"
Let's assume `array = [ 0 ]`. Should `array.pop()` result in `{ op: "remove", path: [ "0" ] }` instead of `{ op: "replace", path: [ "length" ], value: 0 }`?

I guess Immer is already "breaking" from [the spec](https://tools.ietf.org/html/rfc6902) by making `path` an array and not `/foo/bar`, but you could argue replacing `length` isn't friendly to non-JS environments.

I assume the current behavior improves performance slightly.

https://codesandbox.io/s/0o1zx5226l

  * Version: **1.7.2**
  * [X] Occurs when using Proxies (use `setUseProxies(true)`)
  * [X] Occurs in the ES5 implementation (use `setUseProxies(false)`)

jsdevtom:
Fixes #208
Please look for things I haven't tested. I found the part of the code that I was reading very hard to understand due to the lack of descriptive names and unexpected swapping of those variable names that I did understand (see https://github.com/immerjs/immer/blob/master/src/patches.js#L15). For this reason I am not sure that I have not introduced unexpected behavior.

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -249,9 +249,10 @@ describe("arrays - modify and shrink", () => {
 			d.x[0] = 4
 			d.x.length = 2
 		},
+		[{op: "replace", path: ["x", 0], value: 4}, {op: "remove", path: ["x", 2]}],
 		[
-			{op: "replace", path: ["x", 0], value: 4},
-			{op: "replace", path: ["x", "length"], value: 2}
+			{op: "replace", path: ["x", 0], value: 1},
+			{op: "add", path: ["x", 2], value: 3}
 		]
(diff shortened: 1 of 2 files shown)


mweststrate:
Looks great at first sight! Will study it in more detail when cutting next release

mfix22:
Such a cool PR! Fixing bugs by _deleting_ code. Looks like the right move 🎉

mweststrate:
Looks correct to me!

Issue #260, opened 2018-12-03
Array values of produced Object are not frozen? Limitation or bug?


suitedjimster:
* [ ] Issue: Passing an array of objects, does not seem to deep freeze deep arrays.
  * Version: 1.8.0
  * _SIMPLE_ Reproduction: https://codesandbox.io/s/oqnrqlorj9
  * Expected behavior: All deep children object including arrays would be frozen.
  * Actual behavior:  Arrays inside an arrays of objects are not frozen and are allowed to be mutated outside of the produce function

mweststrate:
Immer currently doesn't apply freezing to the full tree, but only to the parts which are touched by immer itself. Full tree freezing might be expensive and can be easily achieved outside immer (for example by using the `deepfreeze` package.

aleclarson:
Objects you create in a producer are skipped by Immer during the finalization phase. You must be careful not to deep-freeze an object that contains a draft. If you're not sure, deep-freeze the object _outside_ its producer, since then you can guarantee any drafts have been swapped out.

Related: #230

aleclarson:
The above solution can now be used in v1.9.0+

Issue #230, opened 2018-10-25
Nested drafts inside new objects


aleclarson:
This scenario seems like it's handled wrong by Immer.

Imagine creating a new object in your producer and adding a draft object as a property.

If your new object is ever frozen before your producer returns, you end up with the draft object never being replaced with the original. And worse, you now have a revoked proxy instead of the object you wanted.

I think Immer should always clone frozen objects when any of their properties is a draft object.

Here's a failing test case:

```js
it("ensures new objects never contain a draft", () => {
    const base = {a: {}}
    const next = produce(base, draft => {
        return {
            foo: {a: draft.a},
            bar: Object.freeze({a: draft.a})
        }
    })

    // Mutable object
    expect(original(next.foo.a)).toBeUndefined()
    expect(next.foo.a).toBe(base.a)

    // Frozen object
    expect(original(next.bar.a)).toBeUndefined()
    expect(next.bar.a).toBe(base.a)
})
```

mweststrate:
If you are deliberately force freezing something (why would you), that is just a sign that you want to have literally that thing. So if you really want to forcefully return a draft, then you'll have it, despite it begin useless. I think this use case is no different then: 

```javascript
produce(base, draft =>({
   foo: {a: draft.a},
   bar() { return draft }
}).bar().a // throws, you leaked a draft
```

The goal is to behave good if you have good intentions. Freezing a draft looks to me like deliberately trying to counter the abstraction provided (namely, having mutable, not frozen drafts), and it that case you are on your own :)

aleclarson:
There's also a chance that a user wants to `deepFreeze` some new object that happens to contain a draft.

aleclarson:
What if we provide a `freeze` function that prevents draft leaks?

aleclarson:
*Note:* To email users, I made edits to past comments.

@mweststrate So you're asking _every producer_ to use `original` before passing a draft object to _any function_ whose return value is ever (1) _frozen_ and (2) added to the draft / replaces the draft. Is this an accurate depiction?

aleclarson:
ping @mweststrate

mweststrate:
I think that introducing a freeze utility + the fact that immer freezes itself will be highly confusing. I would rather propose a more aggressive setting like `setAutoFreeze("everything")`, which is more expensive, but always recursively freezes stuff

[...]

PR #417, opened 2019-09-10
BREAKING CHANGE: In development mode, any new value stored in a tree will be deeply frozen

### LINKED ISSUE #313: Initial state immutability
I am currently exploring immer for a use case of improving an component performance with immutability and angular OnPush change detection. I encountered a strange thing with the auto freezing (which by documentation is as expected) and wanted to know if this is a normal approach or i am pushing to make something work which is not the approach expected by immer.

So i have a complex component state (deeply nested). The sub properties of the object are not created as immutable using the produce but by simple object notation. When produce is called on the state, and the property values are set, immer will only freeze the setter/getter but not the value (which is an object).

I wanted to know if this is something that can be improved by immer or i am tackling this problem wrong and should create every object as immerable or using produce?

### LINKED ISSUE #412: Clarify that the first result from the producer is not immutable unless a change is made
## 🚀 Feature Proposal

As expected, when the producer is first applied to an object (or array) and a change is made, an immutable object results. However, if no such change is made, the result is mutable.

## Motivation

Docs are unclear on this matter. I suggest that it would be better to make the result immutable regardless. But if that doesn't work, it would be good for the docs to be clear on this matter.

With the current implementation, I'm forced to make an arbitrary change to the object to make sure that it is immutable immediately.

## Example

```
    it("Simple object becomes immutable as the produce() call made a change", () => {
        const c = {id: 3};
        const immutable = produce(c, draft => {
            draft.id = 4;
        });
        expect(c.id).eql(3);
        expect(immutable.id).eql(4);
        expect(() => (immutable as any).id = 5).to.throw("");
    });

    it("Simple object is not made immutable as the produce() call made no change", () => {
        const c = {id: 3};
        const immutable = produce(c, draft => {
        });
        (immutable as any).id = 4;
        expect(immutable.id).eql(4);
    });
```

mweststrate:
* [x] Fix #412 #363  #260 #230 
* [x] probably also fix #313, #229 as we will have now side effects during the execution of recipes anyway
* [x] document the side effect with autofreeze enabled!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -548,6 +548,8 @@ isDraft(nextState) // => false
 
 Immer automatically freezes any state trees that are modified using `produce`. This protects against accidental modifications of the state tree outside of a producer. This comes with a performance impact, so it is recommended to disable this option in production. It is by default enabled. By default, it is turned on during local development and turned off in production. Use `setAutoFreeze(true / false)` to explicitly turn this feature on or off.
 
+_⚠️ If auto freezing is enabled, recipes are not entirely side-effect free: Any plain object or array that ends up in the produced result, will be frozen, even when these objects were not frozen before the start of the producer! ⚠️_
+
 ## Immer on older JavaScript environments?
 
 By default `produce` tries to use proxies for optimal performance. However, on older JavaScript engines `Proxy` is not available. For example, when running Microsoft Internet Explorer or React Native (< v0.59) on Android. In such cases, Immer will fallback to an ES5 compatible implementation which works identical, but is a bit slower.
--- a/src/common.js
+++ b/src/common.js
@@ -122,10 +122,8 @@ export function clone(obj) {
 }
 
 export function deepFreeze(obj) {
-	if (!isDraftable(obj)) return
-	if (Object.isFrozen(obj)) return
-	if (isDraft(obj)) return
+	if (!isDraftable(obj) || isDraft(obj) || Object.isFrozen(obj)) return
 	Object.freeze(obj)
 	if (Array.isArray(obj)) obj.forEach(deepFreeze)
-	for (const key in obj) deepFreeze(obj[key])
+	else for (const key in obj) deepFreeze(obj[key])
 }
--- a/src/common.js
+++ b/src/common.js
@@ -120,3 +120,12 @@ export function clone(obj) {
 	for (const key in obj) cloned[key] = clone(obj[key])
 	return cloned
 }
+
+export function deepFreeze(obj) {
+	if (!isDraftable(obj)) return
+	if (Object.isFrozen(obj)) return
+	if (isDraft(obj)) return
+	Object.freeze(obj)
+	if (Array.isArray(obj)) obj.forEach(deepFreeze)
+	for (const key in obj) deepFreeze(obj[key])
+}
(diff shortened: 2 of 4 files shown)


ianstormtaylor:
This is awesome, thank you @mweststrate!

src/immer.js
```diff
@@ -299,6 +304,8 @@ export class Immer {
 			// Search new objects for unfinalized drafts. Frozen objects should never contain drafts.
 			else if (isDraftable(value) && !Object.isFrozen(value)) {
 				each(value, finalizeProperty)
+				// for externally incoming object trees, we wan't to make sure they're frozen automatically
```
aleclarson:
```suggestion
				// Freeze new objects when "autoFreeze" is enabled.
```

readme.md
```diff
@@ -548,6 +548,8 @@ isDraft(nextState) // => false
 
 Immer automatically freezes any state trees that are modified using `produce`. This protects against accidental modifications of the state tree outside of a producer. This comes with a performance impact, so it is recommended to disable this option in production. It is by default enabled. By default, it is turned on during local development and turned off in production. Use `setAutoFreeze(true / false)` to explicitly turn this feature on or off.
 
+_⚠️ If auto freezing is enabled, recipes are not entirely side-effect free: Any plain object or array that ends up in the produced result, will be frozen, even when these objects were not frozen before the start of the producer! ⚠️_
```
aleclarson:
```suggestion
_⚠️ If auto freezing is enabled, recipes are not entirely side-effect free: Any plain object or array that ends up in the produced result will be frozen, even when these objects were not frozen before the start of the producer! ⚠️_
```

src/common.js
```diff
@@ -120,3 +120,12 @@ export function clone(obj) {
...
 	return cloned
 }
+
+export function deepFreeze(obj) {
+	if (!isDraftable(obj)) return
+	if (Object.isFrozen(obj)) return
+	if (isDraft(obj)) return
+	Object.freeze(obj)
+	if (Array.isArray(obj)) obj.forEach(deepFreeze)
+	for (const key in obj) deepFreeze(obj[key])
```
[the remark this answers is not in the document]
mweststrate: nope, nice catch!

src/immer.js
mweststrate: Removed the comment, didn't really ad anything

src/immer.js
```diff
@@ -87,8 +88,10 @@ export class Immer {
 			return this.processResult(result, scope)
 		} else {
 			result = recipe(base)
-			if (result === undefined) return base
-			return result !== NOTHING ? result : undefined
+			if (result === NOTHING) return undefined
+			if (result === undefined) result = base
+			this.maybeFreeze(result, true)
+			return result
```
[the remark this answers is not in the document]
mweststrate: I think it is just github displaying tabs to be so big. Prettier seems quite happy with it at tabs=2

mweststrate:
Jeez is semantic release so fragile. This didn't trigger a release, can't really see why....

mweststrate:
@aleclarson how to trigger a major release? I followed the manual, but it doesn't recognize the commit?  Any ideas?

mweststrate:
wow, that was burried deep

aleclarson:
You have to put `BREAKING CHANGE` at the end of the commit body

mweststrate:
jeez, I guess my screen contrast is not high enough to deduct that from their readme ... :/

PR #422, opened 2019-09-13
Docs: remove conflicting desc. of default freezing behavior

xkxx:
Hi Immer maintainers,

Just a small drive-by docs fix! This PR clarifies the default freezing behavior by removing one of the conflicting descriptions.

Thanks,

xkxx


mweststrate:
Nice!

PR #426, opened 2019-09-16
Update example-reducer.md

joe-denea:
- update the standard Redux example to more closely mirror the logic in the Immer reducer
- improve wording in description
- remove codefund "fallback content" comment since there is no fallback content

### DIFF
--- a/docs/example-reducer.md
+++ b/docs/example-reducer.md
@@ -3,30 +3,27 @@ id: example-reducer
 title: Example Reducer
 ---
 
-<div id="codefund"><!-- fallback content --></div>
+<div id="codefund"></div>
 
-Here is a simple example of the difference that Immer could make in practice.
+Here is a simple example of the difference that Immer can make in practice.
 
 ```javascript
 // Redux reducer
-// Shortened, based on: https://github.com/reactjs/redux/blob/master/examples/shopping-cart/src/reducers/products.js
+// Modified for brevity. Based on: https://github.com/reactjs/redux/blob/master/examples/shopping-cart/src/reducers/products.js
 const byId = (state, action) => {
 	switch (action.type) {
 		case RECEIVE_PRODUCTS:
-			return {
-				...state,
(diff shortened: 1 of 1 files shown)


docs/example-reducer.md
```diff
@@ -3,30 +3,27 @@ id: example-reducer
...
+Here is a simple example of the difference that Immer can make in practice.
 
 ```javascript
 // Redux reducer
-// Shortened, based on: https://github.com/reactjs/redux/blob/master/examples/shopping-cart/src/reducers/products.js
+// Modified for brevity. Based on: https://github.com/reactjs/redux/blob/master/examples/shopping-cart/src/reducers/products.js
 const byId = (state, action) => {
 	switch (action.type) {
 		case RECEIVE_PRODUCTS:
-			return {
```
mweststrate: I'd rather leave this one as is, as that is how it is promoted in the Redux docs. This solution is as far as I can see indeed equivalent, however, these solutions are often general patterns, which might become 'unrecognizable' to some people when deviating. So I'd rather keep this one aligned with the Redux docs. (although to be fair, this is example is a stretch anyway, maybe there is a better one in there).

joe-denea: I looked over the examples and a lot of the code seems fairly dubious. I can't find anything written in a standardized fashion like what someone familiar with Redux might be used to, since almost all of the reducers are storing an individual state element like a list of TODOs or a map of products like we have here, which can make them a bit less extensible since it becomes harder to simultaneously update properties containing metadata associated with the state element if any are added in the future without performing a major rewrite. This example is among the closest to the usual patterns, but only because it was written to look that way despite still doing the same thing as the other stores, which is what makes the code so strange. It might be easier just to make a simple example where a single state property is set to an arbitrary value just to demonstrate the standard style.

joe-denea: I tried updating the example to a very simple, standard looking reducer not based off of one of the Redux examples. Let me know if you think this is clearer.

PR #436, opened 2019-10-04
Use legacy proxies if Proxy.revocable is not defined. Fixes #435

### LINKED ISSUE #435: Immer breaks in browsers with Proxy support if they don't have Proxy.revocable
## 🐛 Bug Report

Immer does not work in browser with support for `Proxy`, but not `Proxy.revocable`, for example Firefox 24 (which we unfortunately need to support).

Firefox 24 reports:

```
[10:05:11.037] TypeError: Proxy.revocable is not a function @ http://10.151.32.17:5000/static/js/2.078cae13.chunk.js:1
```

## To Reproduce

- `git clone git@github.com:esturcke/bug-immer-ff24.git`
- `yarn build`
- `serve -s build`
- Access the page in Firefox 24

## Expected behavior

The page should render "Hello Immer".

## Link to repro

https://github.com/esturcke/bug-immer-ff24

## Environment

-   **Immer version: 4.0.0**
-   [x] Occurs with `setUseProxies(true)`
-   [ ] Occurs with `setUseProxies(false)` (ES5 only)

## Potential Fix

Maybe fall back in the non-Proxy version if `Proxy.revocable` is undefined:

```diff
@@ -19,7 +19,7 @@ import {ImmerScope} from "./scope"
 function verifyMinified() {}

 const configDefaults = {
-	useProxies: typeof Proxy !== "undefined" && typeof Reflect !== "undefined",
+	useProxies: typeof Proxy !== "undefined" && typeof Proxy.revocable !== "undefined" && typeof Reflect !== "undefined",
 	autoFreeze:
 		typeof process !== "undefined"
 			? process.env.NODE_ENV !== "production"
```

esturcke:
Some older browsers including Firefox 24 have support for `Proxy`, but not `Proxy.revocable`. This ensures that we fall back on the legacy browser behavior when `Proxy.revocable` is missing.

### DIFF
--- a/src/immer.js
+++ b/src/immer.js
@@ -19,7 +19,10 @@ import {ImmerScope} from "./scope"
 function verifyMinified() {}
 
 const configDefaults = {
-	useProxies: typeof Proxy !== "undefined" && typeof Reflect !== "undefined",
+	useProxies:
+		typeof Proxy !== "undefined" &&
+		typeof Proxy.revocable !== "undefined" &&
+		typeof Reflect !== "undefined",
 	autoFreeze:
 		typeof process !== "undefined"
 			? process.env.NODE_ENV !== "production"


mweststrate:
Nice catch, thanks!

PR #437, opened 2019-10-04
es5 support for Map/Set

runnez:
Hi guys! This is a draft solution for es5 map support. What do you think about this approach? 
It should works with babel preset, also it may works in IE11 ;) 

Now `map draft` tests passes and I'm planning to add Set support, patches, etc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
     - "8"
     - "10"
-    - "node"
+    # - "node"
 env:
     - NODE_ENV=TEST
 cache:
--- a/__tests__/__snapshots__/hooks.js.snap
+++ b/__tests__/__snapshots__/hooks.js.snap
@@ -1,5 +1,90 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`hooks (es5) - onAssign() when draft is a Map assign 1`] = `
+Array [
+  Array [
+    "a",
+    10,
+  ],
+  Array [
+    Object {
+      "prop": "val1",
+    },
+    11,
+  ],
+]
+`;
+
+exports[`hooks (es5) - onAssign() when draft is a Map nested assignments 1`] = `
+Array [
+  Array [
+    "b",
+    2,
+  ],
+  Array [
+    Object {
+      "prop": "val1",
+    },
+    Map {
+      "b" => 2,
+      "d" => 1,
+    },
+  ],
+  Array [
+    "a",
+    Map {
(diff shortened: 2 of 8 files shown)


runnez:
Hmm, CI failed, I guess according this https://github.com/immerjs/immer/commit/017de43395044faa703c22eef441aa3b717a42d7

mweststrate:
Awesome! I'll hope to review in detail hopefully somewhere next week

runnez:
Need to decide
- Should it work in environment without Map and Set? 
If yes, I need to add a checking 
- Should it work in IE11 without babel? 
IE11 has partial support without iteration methods and difference between `Map.set` method and does not support `new Map(iterable)`

And I have doubts about a lot of copy-paste between implementations proxy and es5

mweststrate:
1. No, as they can be polyfilled, which is the responsibility of the consumer of this package. A nice error message about the need of this would be great though.
2. I think it is fine sticking to the normal web api's, and if something doesn't work on any browser, it is up to the host to set up some polyfill. So I'd write the code as if it is targetting a modern browser. 
3. I can imagine that having only an ES5 implementation might work, that is, a class based approach, and use that one in the Proxy case as well? I mean, the fact that Proxies are available, doesn't mean that we must use them :). I'd opt for code-base maintainability here.

runnez:
> No, as they can be polyfilled, which is the responsibility of the consumer of this package. A nice error message about the need of this would be great though.

We have some checking here 
https://github.com/immerjs/immer/pull/437/files#diff-de99f837249e5adf62716789c7b9e80eR40
Imagine, some consumer no uses Map or Set if he updates Immer, it will fail in ES5 browsers without Map and Set

runnez:
@mweststrate I think I've done. But I don't understand one thing. Should I update `markChangesRecursively`? It skips `Map` and `Set`, should it be an explicit?

mweststrate:
Awesome! I don't know the answer to that question from the top of my head,
but if a test in which you modify an entry in a map, that is in a map, that
is in an object, and that works, it's correct, as the mechanism is used to
ensure that changes propagate through unmodified parents. I'm gonna try to
carve out some time tomorrow to review in detail!

mweststrate:
Superficially this looks great! I'll merge it into the v4 branch, and to further local investigation there. I'll also send a maintainer invite. That will allow you to directly commit to that branch, which avoids the whole song-and-dance of maintaining your own fork :)

runnez:
@mweststrate Glad to hear that! Thanks a lot for invitation!

Issue #392, opened 2019-06-26
Skip key in shallowCopy


zewa666:
## 🚀 Feature Proposal

It would be great if the user could control the shallowCopy in a way to tell what keys to skip during the process. The easiest thing would be if we could provide a predicate which returns bool at this [specific part](https://github.com/immerjs/immer/blob/master/src/common.js#L60) providing it the `key` and `base`.

## Motivation

One of the possible use-cases for this might be for Frameworks such as [Aurelia](https://aurelia.io) and it's state management plugin [Aurelia Store](https://github.com/aurelia/store) to by-pass the built in observable trackers placed on objects by Aurelia.

## Example

I've no idea how to register the predicate in order to consume it in the [commons helper](https://github.com/immerjs/immer/blob/master/src/common.js#L60) but maybe something along these lines:

```javascript
import { setSkipCopyPredicate } from "immer";

function myPredicate(base, key) {
  return (key !== "FOOBAR")
}

setSkipCopyPredicate(myPredicate);
```

aleclarson:
Help me understand the issue better. Why would you want to bypass the "observable trackers" added by Aurelia? Please provide some example code as well.

zewa666:
the issue with those is that they are computed getter/setters, and out of the box throw an error with Immer.js which pretty much makes Immer unusable with Aurelia. Same as it seems with Vue and Ember as described [here](https://github.com/immerjs/immer/issues/317)

zewa666:
@aleclarson, is there anything else I can do?

aleclarson:
Are you sure you want to skip computed properties? That seems like it would break the observability that Aurelia provides, but that might make sense depending on your use case (I've never used Aurelia). Could you setup a CodeSandbox that demonstrates how you're using Immer with Aurelia?

zewa666:
This is not about Aurelia itself but its state management plugin called Aurelia Store. So its for a pretty narrowed down use case where people are still able to use alternative ways like Object.assign spreads and so on.
Moreover we'll be looking into means to address this in future with a core solution. But for now the current behavior of throwing errors on computeds makes Immer pretty much unusable for Aurelia and so it seems also other frameworks.

There might also be other use cases, such as strippin computeds altogether if the resulting object e.g is going to be serialized in the end (localstorage, Redux devtools, ...)

mweststrate:
[...]

Issue #317, opened 2019-02-15
Why can't drafts have computed properties?


Krantisinh:
**Using immer v1.9.0+ with VueJS gives error 'Immer drafts cannot have computed properties'**

* [x] Issue: _______
  * Version: 1.9.0 onwards
  * Expected behavior: Updating state using immer produce function runs without errors.
  * Observed behavior: Immer (1.9.0+), when used with VueJS (2.5.0+) throws an error  'Immer drafts cannot have computed properties' while updating the state.

On further analysis, it looks like -
Vue JS sets getters in all properties used in its components by default. 
And when we looked at the source of immer, we found that, below code throws error once it encounters a getter on any prop while making a shallow copy.

Immer is an awesome library and we want to continue using the same, but this issue is blocking us. Could you please explain the rationale behind below change and let us know how we can help to fix this issue ?

```
export function shallowCopy(base, invokeGetters = false) {
    if (Array.isArray(base)) return base.slice()
    const clone = Object.create(Object.getPrototypeOf(base))
    ownKeys(base).forEach(key => {
        if (key === DRAFT_STATE) {
            return // Never copy over draft state.
        }
        const desc = Object.getOwnPropertyDescriptor(base, key)
        if (desc.get) {
            if (!invokeGetters) {
                throw new Error("Immer drafts cannot have computed properties")
            }
            desc.value = desc.get.call(base)
        }
        if (desc.enumerable) {
            clone[key] = desc.value
        } else {
            Object.defineProperty(clone, key, {
                value: desc.value,
                writable: true,
                configurable: true
            })
        }
    })
    return clone
}
```

mweststrate:
@Krantisinh why are you using immer with VueJS objects? VueJS are designed to be mutable so that they can be tracked by Vue, and Immer is trying to achieve to do exactly the opposite, so using immer on these kind of objects doesn't really add any value?

The problem with getters is that they can't be cloned. What should be cloned? The value? But in that case they become 'static'. Or should the descriptor be copied? But in that case the closure of the getter might be wrong, that is, referring to something it was already referring to, there is no way to tell..

aleclarson:
Note: You can still have computed properties on the **prototype** of your objects.

oriSomething:
I think Immer is a valid choice with Vue in some cases. But, I think the issue is with Vue and not Immer, since using Vue is a package deal of Vue modify your objects.

Krantisinh:
[...]

PR #438, opened 2019-10-06
Fix: Cannot set property of  which has only a getter

dalcib:
When a method of a Class makes changes in a cloned copy of a getter property wich type is an array, the property is included in the drafts, causing the TypeError: Cannot set property 'propName' of #<Class> which has only a getter. For example:
```javascript
class State {
  [immerable] = true
  word1 = 'bar'
  word2 = 'foo'
  get chars() {
    return this.word1.concat(this.word2).split('')
  }
  setWord(newWord) {
    let mix = [...this.chars].slice(0,3)
    // The next line changes a cloned copy of this.chars
    mix[2] = 'z'
    this.word1 = mix.slice(0, 3).join('')
    this.word2 = newWord
  }
}

const state = new State()

test('should allow getter array', () => {
  expect(state.word1).toEqual('bar')
  expect(state.chars).toEqual(['b', 'a', 'r', 'f', 'o', 'o'])
  const newState = produce(state, d => d.setWord())  //TypeError: Cannot set property chars of #<State> which has only a getter
  expect(newState.word2).toEqual('baz')
})
```

This PR fixes it, checking with `ownKeys` if the property exists in the state.base, before creating the proxy. If the property doesn't exist, it returns the value.

https://codesandbox.io/s/great-colden-1py9k


dalcib:
This PR also solves #439

mweststrate:
Thanks for the suggested fix! Could you also include a unit test that makes sure we don't regress on this and the code is covered? Thanks!

dalcib:
The tests were included @mweststrate. Thanks.

__tests__/base.js
```diff
@@ -1147,6 +1147,47 @@ function runBaseTest(name, useProxies, autoFreeze, useListener) {
...
+		const state = new State()
+
+		it("should work without creating a proxy for a getter property", () => {
+			expect(state.chars).toEqual(['b', 'a', 'r'])
+			const newState1 = produce(state, d => d.setWord2())
+			expect(newState1.word2).toEqual('baz')
+
+			expect(state.bar).toEqual({ baz: 1 })
+			const newState2 = state.syncFoo()
+			expect(newState2.foo).toEqual(1)
```
mweststrate: I'd check here as well if `newState2.bar` output is correct (in a case were it differs from newState1), to verify that `this.` works everywhere as expected in those getters

src/proxy.js
```diff
@@ -130,6 +131,10 @@ function get(state, prop) {
 		// Store drafts on the copy (when one exists).
 		drafts = state.copy
 	}
+	
+	if (!ownKeys(state.base).includes(prop)) {
```
mweststrate: Note that you are checking for enumerability here, not for getter-ness. Those two are in principle unrelated, excepti that getters are by default non-enumerable. 

To find getters and setters you need check `Object.getOwnPropertyDescriptor(object, propname)`, and see if that has a `get` or `set` method.

src/proxy.js
```diff
@@ -131,6 +131,10 @@ function get(state, prop) {
 		drafts = state.copy
 	}
 
+	if (!Object.keys(state.base).includes(prop)) {
```
dalcib: `Object.getOwnPropertyDescriptor(state.copy || state.base, prop)` always returns `undefined` for getters and methods. So, to fix the problem, it's enough to check where the property exist or not in the state.

__tests__/base.js
dalcib: Done

dalcib:
Hi @mweststrate. The use of getOwnPropertyDescriptor didn't work, because it always returns undefined for getters. I am proposing to use Object.keys() instead.

mweststrate:
Ah, that is because they probably live on the prototype, not the instance. I think we should hold this a little, and think this further through or maybe even forbid it. This is a rabbit hole that might go quite deep, see #317 (and #392) for some background. 

I am not sure there is an universally applicable solution to this.

dalcib:
@mweststrate, now that v. 5.0.0 is ready, can we think about this issue?
This problem only occurs with `setUseProxies(true)`.  With es5 (without proxies) the getters work nicely. You can check it out at https://codesandbox.io/s/useimmer-tic-tac-ci6qq.
There is no performance or bundle size cost in the fix.
It allows the same behavior, with and without proxies.
@luncheon

src/proxy.js
```diff
@@ -107,6 +107,14 @@ const objectTraps = {
...
 			drafts = state.copy
 		}
+		
+		if (!Object.keys(state.base).includes(prop)) {
+			return value
+		}
+
+		if (!Object.keys(state.base).includes(prop)) {
+			return value
+		}
```
Venryx: Are my eyes deceiving me, or are there two exact copies of the keys-check + return code? (typo?)

dalcib: Thanks

dalcib:
Replaced by https://github.com/immerjs/immer/pull/558

PR #440, opened 2019-10-08
remove non-existent import

henryqdineen:
I noticed this errant import while working on this repo locally. It looks like it causes rollup/bili warnings. 

```
warning circular_dependency: Circular dependency: src/immer.js -> src/patches.js -> src/immer.js
warning non_existent_export: Non-existent export 'createDraft' is imported from src/immer.js
```


mweststrate:
Nice catch, thanks!

PR #473, opened 2019-12-03
chore: use SVG logo

slikts:
Replaces the logo with an SVG so that it scales better and doesn't have aliased edges on a dark background (like for Dark Reader users).


slikts:
I've added PNG versions without aliasing.

PR #474, opened 2019-12-04
Fixing #472 (In proxy, Freezed Maps don't become draft)

### LINKED ISSUE #472: Unexpected access to original frozen objects while updating a map
## 🐛 Bug Report

When updating several objects in a `Map`, writing to an object that should be mutable throws an error in some cases.

## To Reproduce

This code reproduces the error.
```
import {produce} from "immer"

const project = produce(new Map(), draft => {
  draft.set("bar1", {blocked: false})
  draft.set("bar2", {blocked: false})
})

// Read before write -- no error
produce(project, draft => {
  const bar1 = draft.get("bar1")
  const bar2 = draft.get("bar2")
  bar1.blocked = true
  bar2.blocked = true
})

// Read/write interleaved -- error
produce(project, draft => {
  const bar1 = draft.get("bar1")
  bar1.blocked = true
  const bar2 = draft.get("bar2")
  bar2.blocked = true // TypeError: "blocked" is read-only
}) 
```

## Expected behavior

Writing to objects inside a map in `produce` should not error. More broadly it seems like accessing frozen objects in `produce` should only be possible with `original`.

## Link to repro (highly encouraged)

[CodeSandbox demo](https://codesandbox.io/s/immer-sandbox-84tkm)

## Environment

- **Immer version: 5.0.0**

oriSomething:
This PR fixes #472, which in cases `Map`s throw. The problem that the test was pass before fix, even though in [the browser](https://codesandbox.io/s/icy-frog-flbdu) it would throw, not sure why.

Also, even tough it fixes the issue, I'm not sure this the root of problem. This solution might be only a patch.

Help is welcome in navigating Immer

### DIFF
--- a/__tests__/__snapshots__/manual.js.snap
+++ b/__tests__/__snapshots__/manual.js.snap
@@ -50,7 +50,7 @@ Array [
 ]
 `;
 
-exports[`manual - proxy cannot modify after finish 1`] = `"Cannot use a proxy that has been revoked. Did you pass an object from inside an immer function to an async process? {\\"a\\":2}"`;
+exports[`manual - proxy cannot modify after finish 1`] = `"Cannot perform 'set' on a proxy that has been revoked"`;
 
 exports[`manual - proxy should check arguments 1`] = `"First argument to \`createDraft\` must be a plain object, an array, or an immerable object"`;
 
@@ -60,7 +60,7 @@ exports[`manual - proxy should check arguments 3`] = `"First argument to \`finis
 
 exports[`manual - proxy should not finish drafts from produce 1`] = `"First argument to \`finishDraft\` must be a draft returned by \`createDraft\`"`;
 
-exports[`manual - proxy should not finish twice 1`] = `"The given draft is already finalized"`;
+exports[`manual - proxy should not finish twice 1`] = `"Cannot perform 'get' on a proxy that has been revoked"`;
 
 exports[`manual - proxy should support patches drafts 1`] = `
 Array [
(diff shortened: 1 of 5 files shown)


oriSomething:
1. The solution seems to be correct
2. But proxies tests for some reason executes in ES5 commands, which made the test before fix pass.

oriSomething:
After fixing broken tests there is a big issue with proxies:
The test of `combines with produce - 2` is broken because proxies are broken.
When a new draft created it takes the draft property as is, and if it gets modified, it affect the new proxy, I guess there are some solutions and I need your thoughts about how to fix this (@mweststrate , @aleclarson):

1. ~~Copy copied state when created~~ bad idea because if the base draft won't be modified, it will break `{} = {}`
2. Modified base draft so `base` become `copy` (Probably more changes are needed) and than by then if the base draft changes it won't break the created draft from this draft. (Maybe it make unwanted side effect)?
3. Some hook that get applied when base draft gets modified, and clone object before change, but I think it will be complex and buggy
4. Any other solutions?

mweststrate:
@oriSomething yeah I think there is some iffy stuff in there. I'm gonna try to take a closer look somewhere this or next week. Probably unifying it to the non-proxy implementation first, I think the proxy map complicates things more than it helps.

oriSomething:
Just not, the problem that left isn't `Map` related, it's related to how drafts in proxies work

mweststrate:
Ok sorry! Need to read more carefully I guess 😊

Op do 5 dec. 2019 17:50 schreef Ori Livni <notifications@github.com>:

mweststrate:
Thanks! REleasing this one, although it doesn't solve everything, it solves some cases. For follow up see #478

Issue #466, opened 2019-11-21
Map and Set.Patches bug and error throws.


tinder1024:
## 🐛 Bug Report

1.Set can not record patches.
2.Map can not record patches some times.
3.Map throw an exception when calling mutative inside a producer.

## To Reproduce

```javascript
const { produce, createDraft, finishDraft } = require('immer');

/**
 * Set can not record patches
 */
const setNoPatches = function(){
    const obj = {
        set:new Set()
    }

    const result = produce(obj,draft=>{
        draft.set.add('abc');
    },(patches,inversePatches)=>{
        //patches and inversePatches is empty array
        console.log('PATCHES:',patches,inversePatches);
    });
    console.log('RESULT:',result);
}


/**
 * Map can not record some patches
 */
const mapChangeBug = function(){
    const obj = {
        map:new Map([
            ['a',new Map([
                ['b',true],
                ['c',true],
                ['d',true],
            ])],
            ['b',new Map([
                ['a',true]
            ])],
            ['c',new Map([
                ['a',true]
            ])],
            ['d',new Map([
                ['a',true]
            ])],
        ])
    };
    const result = produce(obj,draft=>{
        const aMap = draft.map.get('a');
        aMap.forEach((_,other)=>{
            const otherMap = draft.map.get(other);
            otherMap.delete('a');
        })
    } ,(patches,inversePatches)=>{
       //no patches of map 'b' 'c' 'd'.
        console.log('PATCH:',patches,inversePatches);
    });
    console.log('RESULT',result);
}
/**
 * Error:This object has been frozen and should not be mutated
 */
const mapChangeBug2 = function(){
    const obj = {
        map:new Map([
            ['a',new Map([
                ['b',true],
                ['c',true],
                ['d',true],
            ])],
            ['b',new Map([
                ['a',true]
            ])],
            ['c',new Map([
                ['a',true]
            ])],
            ['d',new Map([
                ['a',true]
            ])],
        ])
    };
    const obj1 = produce(obj,draft => {});
    const result = produce(obj1,draft=>{
        const aMap = draft.map.get('a');
        aMap.forEach((_,other)=>{
            const otherMap = draft.map.get(other);
            otherMap.delete('a');
        })
    } ,(patches,inversePatches)=>{
        console.log('PATCH:',patches,inversePatches);
    });
    console.log('RESULT',result);
}

setNoPatches();
mapChangeBug1();
//mapChangeBug2();
```

## Expected behavior

[...]

Issue #449, opened 2019-10-30
Reduce Immer bundle size by eliminating the Proxy based implementations of Set and Map


mweststrate:
Currently all data structures supported by Immer (plain objects, arrays, Maps, Sets) have an ES5 and Proxy based implementation. However, the Proxy implementation of Maps and Sets don't really add value, as all object interactions are through patchable methods, so the ES5 implementation has no real practical or performance drawback over the Proxy implementation. Eliminating the latter should reduce the bundle size with 0.5 to 1 KB? Relevant [comment](https://github.com/immerjs/immer/pull/354#discussion_r340535391)

oriSomething:
As much as I don't like this idea (because mutating bulllit-ins) it can solve #466 for free

oriSomething:
By the way, have someone benchmarked if `Object.create(···)` makes slower or faster ES5 `Map`/`Set` traps?

mweststrate:
Released as 5.2.0

PR #478, opened 2019-12-11
Fix maps, kill proxy implementation for Maps

### LINKED ISSUE #472: Unexpected access to original frozen objects while updating a map

mweststrate:
- [x] finish clean code branch first
- [x] test / mention map keys are not drafted (or open issue to do so)
- [x] test / mention external data objects are not draft (or open issue to do so)

Should fix #472, #466, #449


__tests__/map-set.js
```diff
@@ -0,0 +1,176 @@
...
+runBaseTest("proxy (no freeze)", true, false)
+runBaseTest("proxy (autofreeze)", true, true)
+runBaseTest("proxy (autofreeze)(patch listener)", true, true, true)
+runBaseTest("es5 (no freeze)", false, false)
+runBaseTest("es5 (autofreeze)", false, true)
+runBaseTest("es5 (autofreeze)(patch listener)", false, true, true)
+
+function runBaseTest(name, useProxies, autoFreeze, useListener) {
+	const listener = useListener ? function() {} : undefined
+	const {produce, produceWithPatches} = createPatchedImmer({
```
oriSomething: I think to prevent bugs in future, it would be better be inside `beforeEach`. Otherwise, If `new Immer()` isn't 100% clean from global environment it will causes bugs in testing, as was before

JackCA:
Any change we see this go out soon?

mweststrate:
I expect and of next week or early after that. Got most functionality
working again, but the devil is in the details here

runnez:
@mweststrate I could help with something if needed

tsconfig.json
```diff
@@ -2,6 +2,8 @@
   "include": ["src"],
   "compilerOptions": {
     "lib": ["es2015"],
+    // TODO: change to ES5 and make sure maps still work
```
runnez: I tried to test current implementation with extending map (I like it and I'm curious how it works :) 
I found some issue with TS output to ES5 and V8 
```
Uncaught TypeError: Constructor Map requires 'new'
    at DraftMap.Map (<anonymous>)
    at new DraftMap (<anonymous>:19:28)
    at <anonymous>:1:1
```
https://github.com/microsoft/TypeScript/issues/10853

mweststrate: Yeah, that is how far I got :D. Probably, if extending is not possible, we should do a dirty set prototype hack. The most important thing is that `draftMap instanceof Map` passes, beyond that we don't really make / need any super calls

mweststrate: This is probably an interesting thing to figure out if you are interested :) Beyond that it might be a bit hard to help, as I am a bit all over the place, refactoring to reduce code duplication, (or just untangle some spaghetti that grew over time), so it is kinda hard to split that up

PR #484, opened 2019-12-18
TS conversion

mweststrate:


tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "lib": ["es2015"],
     "strict": true,
+    "declaration": true,
```
mweststrate: @egoist any idea why this breaks bili? (see error above in commit message of https://github.com/immerjs/immer/pull/484/commits/4377dcd28f5c846f8478ffb81396418ad5bd4fa2)

mweststrate: Apologies, forgot to bump versions first. Latest version fixes it.

mweststrate:
Can be tried using `5.1.0-beta.1`

Issue #487, opened 2019-12-21
Webpack warns about an erroneous sourcemap path


ivan-aksamentov:
## 🐛 Bug Report

When using immer 5.1.0 and source maps, webpack emits the following warning:

```
WARNING in ./node_modules/immer/dist/immer.module.js
Module Warning (from ./node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../node_modules/tslib/tslib.es6.js': Error: Can't resolve '../node_modules/tslib/tslib.es6.js' in '/home/ia/projects/repro-immer-sourcemaps/node_modules/immer/dist'
 @ ./index.js 1:0-31 2:12-17
 @ multi ./index.js

```

## Link to repro

https://github.com/ivan-aksamentov/repro-immer-sourcemaps

## To Reproduce

```bash
git clone https://github.com/ivan-aksamentov/repro-immer-sourcemaps
cd repro-immer-sourcemaps
yarn install
yarn dev

```

(`yarn dev` is equivalent to
`NODE_ENV=development BABEL_ENV=development webpack`)

## Observed behavior

Webpack will emit the following warning:

```
WARNING in ./node_modules/immer/dist/immer.module.js
Module Warning (from ./node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../node_modules/tslib/tslib.es6.js': Error: Can't resolve '../node_modules/tslib/tslib.es6.js' in '/home/ia/projects/repro-immer-sourcemaps/node_modules/immer/dist'
 @ ./index.js 1:0-31 2:12-17
 @ multi ./index.js

```

The warning is due to incorrect `sources` path entry in
`node_modules/immer/dist/immer.module.js.map`:

```json
{
  "version": 3,
  "file": "immer.module.js",
  "sources": [
    "../node_modules/tslib/tslib.es6.js"
  ],
  ...
}

```

In this case, source map assumes that `tslib.es6.js` is available in
`../node_modules/tslib/tslib.es6.js` (that is inside
`<project_root>/node_modules/immer/node_modules/tslib`), but it may not be the
case, for example if using `yarn` as a package manager, which may hoist
dependencies. In general, source maps should not make such assumptions.

## Expected behavior

Webpack does not emit any warnings

## Environment

```
"os": "Ubuntu 18.04.3 LTS",
"node": '12.14.0',
"yarn": "1.21.1",
"immer": "5.1.0"
[...]

PR #490, opened 2019-12-23
build: Enable sourcemaps and declaration maps in tsconfig.json

### LINKED ISSUE #489: immer source filenames are missing from V5.1.0 sourcemaps
## 🐛 Bug Report

Unfortunately, #467 isn't the only problem with sourcemaps in v5.1.0.  A more serious problem is that the sourcemap doesn't actually map immer source files.  In other words, the v5.1 sourcemap doesn't actually do anything because it doesn't refer to any source files except tslib... which isn't actually a runtime dependency!

The problem is that tsconfig.json needs to enable sourcemaps.  I'm filing a PR shortly to fix this problem.

## Link to repro   / To Reproduce / Observed behavior

Here's the beginning of the 5.1.0 (bad) sourcemap at https://unpkg.com/immer@5.1.0/dist/immer.module.js.map: 
```json
{"version":3,"file":"immer.module.js","sources":["../node_modules/tslib/tslib.es6.js"],"sourcesContent":null,
```
If you look at the rest of the 5.1.0 sourcemap file it's mostly blank, which is expected if there's no immer source files referenced there. 

The user-facing result of the missing mappings is that original source won't be shown in debuggers like Chrome or VSCode.  If users step into immer code, they'll be stepping into the transpiled JS, not the original TS code.

## Expected behavior

Here's the beginning of the (correct) 5.0.2 sourcemap at https://unpkg.com/immer@5.0.2/dist/immer.module.js.map: 

```json
{"version":3,"file":"immer.module.js","sources":["../src/common.js","../src/scope.js","../src/es5.js","../src/proxy.js","../src/patches.js","../src/immer.js","../src/index.js"],"sourcesContent":null,
```

## Environment

- **Immer version:** `5.1.0` (although package.json still says `5.1.0-beta.2` per #488)
- [n/a] Occurs with `setUseProxies(true)`
- [n/a] Occurs with `setUseProxies(false)` (ES5 only)

justingrant:
Fixes #489 by enabling sourcemaps in tsconfig.json. 

While we're in the neighborhood, this PR also enables declaration maps, which are auto-generated sourcemaps for .d.ts files. Declaration maps enable the "Go To Definition" (F12) feature of VSCode, CodeSandbox, and other IDEs to navigate to the actual immer implementation code. Without declaration maps, the user only can only navigate to the .d.ts file when "Go To Definition" is used.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,8 @@
     "strict": true,
     "declaration": true,
     "importHelpers": false,
+    "sourceMap": true,
+    "declarationMap": true,
     "noImplicitAny": false // TODO: true,
   }
 }


mweststrate:
Awesome, thanks! Do you know whether this also fixes the weirdly generated reference to tslib? (See #487)

justingrant:
> Do you know whether this also fixes the weirdly generated reference to tslib? (See #487)

Nope, unfortunately. I couldn't figure out the cause of that tslib thing. See my [comment](https://github.com/immerjs/immer/issues/487#issuecomment-568398708) on #487

Issue #448, opened 2019-10-30
Improving recursive typings when TypeScript 3.7 is out


mweststrate:
_Waiting for TS 3.7 to be released_

The types for `Immutable<T>` and `Draft<T>` are not as accurate as we would love them to be. This is largely caused by TypeScript not yet supporting recursive types. 3.7 should change that, so it would be a good time to revisit these types.

[ ] Improve types as suggested in f8c65f02e3221d35ac6d58dd4f44c40e0df67a55
[ ] Re-evaluate #385
[ ] Fall back to the classic TS recursivity trick?

antonyoneill:
Hey, just a heads up that typescript 3.7 is out 🎉 

https://github.com/microsoft/TypeScript/releases/tag/v3.7.2

PR #495, opened 2020-01-03
Make explicit use of TypeScript's ReadonlySet and ReadonlyMap types. Fixes #494.

### LINKED ISSUE #494: Support Typescript's readonly array, map and set types
## 🚀 Feature Proposal

Typescript has some utility types, ReadonlyArray<V>, ReadonlySet<V> and ReadonlyMap<K, V> but immer doesn't seem to correctly handle these in the Immutable<T> and Draft<T> types.

## Motivation

More consistent type checking is in everyone's best interests!

I could probably write my own typings to override immer's but I would expect this to be something included in immer itself.

## Example

```
import { Draft, produce } from "immer";

let foo: ReadonlySet<string> = new Set();

produce(foo, (draft: Draft<ReadonlySet<string>>) => {
  draft.set("bar");
});
```

Result:

`Property 'set' does not exist on type '{ forEach: (callbackfn: (value: string, value2: string, set: ReadonlySet<string>) => void, thisArg?: any) => void; has: (value: string) => boolean; size: number; entries: { ...; }; keys: { ...; }; values: { ...; }; }'

Mossop:
TypeScript provides the ReadonlySet and ReadonlyMap types which match
their normal counterparts but without any methods that mutate their data.
This change makes Immutable<Map> and Immutable<Set> map to those types
directly.

Worth noting that Map extends ReadonlyMap so we only need to check for
one to know what the resulting type is, the same goes for Set and
ReadonlySet.

This also removes the intermediate Set/Map types since TypeScript no
longer seems to need those (likely since TypeScript 3.7, see #448).

### DIFF
--- a/__tests__/draft.ts
+++ b/__tests__/draft.ts
@@ -187,6 +187,13 @@ test("draft.ts", () => {
 		assert(fromDraft(toDraft(weak)), weak)
 	}
 
+	// ReadonlyMap instance
+	{
+		let val: ReadonlyMap<any, any> = _
+		let draft: Map<any, any> = _
+		assert(toDraft(val), draft)
+	}
+
 	// Set instance
 	{
 		let val: Set<any> = _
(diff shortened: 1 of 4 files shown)


mweststrate:
Thanks for this PR! Merged into #506 and will become part of the next minor

PR #525, opened 2020-02-03
fix in example code of introduction

saurabh-codeword:


mweststrate:
The original code is correct, the base form of produce is calling it with two arguments.

PR #526, opened 2020-02-04
fix: patches when delete object property in array

ryuever:
May be it's about two points:

1. The length is the same, `markChangesRecursively` is still necessary for un-processed array items. such as delete property in an object array, the `nextState` could be correct, but `patches` not works as expected.
2. The `delta` item has been process as `add`/`remove`. The remaining items count would be the `min` value.


mweststrate:
Releasing as 5.3.3. Thanks!

PR #531, opened 2020-02-10
fix: consistent naming for castDraft, fixup 2d27f33f6

denis-sokolov:
The function is called castDraft, the `asDraft` version seems to be a remnant of an earlier name. It’s only present in the documentation and the test titles, so this change is non-functional.

### DIFF
--- a/__tests__/draft.ts
+++ b/__tests__/draft.ts
@@ -304,7 +304,7 @@ test("draft.ts", () => {
 	expect(true).toBe(true)
 })
 
-test("asDraft", () => {
+test("castDraft", () => {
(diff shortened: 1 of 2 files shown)


mweststrate:
Nice catch!

mweststrate:
woops, this should have had tag `docs:` not `fix:`

denis-sokolov:
Apologies, I did not know you use semantic-release bot and I did not put serious effort into figuring out the commit message.

mweststrate:
No worries, I was blaming myself :) That message can be edited when
merging, I just went too fast

PR #536, opened 2020-02-17
feature: optimize Immer bundling

mweststrate:
BREAKING CHANGE: need to enable features

BREAKING CHANGE: custom hooks are no longer supporterd


__performance_tests__/add-data.js
```diff
@@ -1,103 +1,109 @@
 "use strict"
 
 import {measure} from "./measure"
-import produce, {setAutoFreeze, setUseProxies} from "../dist/immer.umd.js"
+import produce, {
+	setAutoFreeze,
+	setUseProxies,
+	enableAllPlugins
+} from "../dist/immer.cjs.production.min.js"
```
FredyC: Do you see some benefit to refer minified prod build in tests? For once it might be harder to run debugger with it. You could simply refer to `../dist` here which will pickup dev build so it would be less verbose.

jest.config.build.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+	moduleNameMapper: {
+		"src/.*": "<rootDir>/dist/immer.esm.js"
```
FredyC: Interesting. Shouldn't you refer to `src` in test files and let it translated to dist with this?

package.json
```diff
@@ -1,27 +1,30 @@
 {
   "name": "immer",
-  "version": "5.3.2",
+  "version": "6.0.0-alpha.2",
   "description": "Create your next immutable state by mutating the current one",
   "main": "dist/immer.js",
```
FredyC:
```suggestion
  "main": "dist/index.js",
```

markerikson:
One question on this: while you listed a couple breaking changes up top, do the main `produce` and `Draft` APIs remain the same?

mweststrate:
yes :)

markerikson:
Cool. That likely means we could bump our Immer dep from 4 to 6 in an RTK minor (say, 1.4.0).

Per my Twitter comment, we don't explicitly need the ES5 aspect ourselves, but it's likely that a number of our users do because they need to run in Proxy-less environments (IE-whatever, RN apparently?).  Depending on how these "plugin" APIs turn out, we might either disable a couple things ourselves, or give instructions to users on importing Immer and flipping those switches as needed.

FredyC:
@mweststrate Have you noticed my review comments above? :)

mweststrate:
yes, will look into it once done :) this is still a draft, just coudn't
find that draft button anymore :P

mweststrate:
@markerikson (and others) the new version can be tried as immer@6.0.0-alpha.3. Installation instructions can be found here: https://github.com/immerjs/immer/blob/multi-bundle/docs/installation.md#pick-your-immer-version

FredyC:
@mweststrate You should definitely fix that `main` field I pointed above, it won't work in any nonESM environment.

__performance_tests__/add-data.js
mweststrate: no, these are the performance test, not the unit tests. So they won't really assert anything and arent to be relied upon for correctness. Using a prod build to better reflect raw production performance

jest.config.build.js
mweststrate: yeah, this makes the tests behave as if immer was a dependency. So if anything is referred from src that is also exported from the index, it is fine, otherwise it will become undefined, so any graybox tests are marked to be not included in this build config

package.json
mweststrate: Nice catch 😅

markerikson:
Can confirm the entry point issue broke things.   I just tried `alpha.3`, and yeah, all our tests blew up with Jest saying it can't find Immer.  Hand-editing `node_modules/immer/package.json` let them run okay.


@mweststrate , can you put out another alpha? (Given the time zone difference, you're probably asleep now, so I'll have to try this again tomorrow night.)

Having said that, I also tried a local test build of RTK with the hand-edited copy of Immer 6, and I can confirm that the size of Immer in a demo app bundle is smaller.

Webpack reports the demo app's `main.js` went from 19.1K down to 11.5K for Immer 6 with no plugins, and 13.5K with the ES5 plugin.

`source-map-explorer` was previously showing Immer as being around 12K by itself, I think. It now shows 6.64K with no plugins and 8.61K with the ES5 plugin.

So, looks like that plugin adds about 2K in this example, but overall, a definite win.

mweststrate:
I'm actually in CA at the moment :) Just pushed an update (didn't test it,
let me know if it isn't ok)

Yeah those numbers align with my expectations for non-gzipped code. I think
the ES5 plugin could actually be pruned a little bit more as there seems to
be some almost-duplication, maybe in a next version.

markerikson:
Yep, `alpha.4` builds cleanly.  Thanks!

Issue #538, opened 2020-02-18
Could you ship a copy of the old type definitions for <3.7 users using typesversions?


phryneas:
## 🚀 Feature Proposal

Right now, it seems you've dropped support for all TS versions below 3.7. I find the prospect of the new typings exciting (we've had our share of problems with the old typings), and I'd like to propose that we bump the `immer` version we're using in `@reduxjs/toolkit` - but our minimum TS version is 3.5 and going up to 3.7 would be very bold and from what I've seen in the issues, it would hurt quite a few people.

So my idea was: could you just ship the old type definitions as a [typesversions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions)-marked `.d.ts`?
Sure, those had some problems, but given the "problematic or nothing" choice, this fallback might benefit quite some people.

From what I've gathered, the actual API hasn't changed since (I might be wrong on that), so this might be just working.

mweststrate:
Feel free to submit a PR with old typings and that field backed in :) I guess @acemark would be quite interested as well

markerikson:
I'm @markerikson here on Github, but yes :)

mweststrate:
Forgot to add to the release notes, but this landed in 6.0.0 as well!

markerikson:
Awesome, thanks!

markerikson:
(btw, you can edit GH release notes any time - goodness knows I've done that frequently)

phryneas:
Sweet, thanks a lot!

PR #541, opened 2020-02-18
add 5.0.2 type definitions as fallback for TS<3.7

phryneas:
As suggested in #538, this should allow TS versions 3.1 (which was the first version to support the `typesVersions` package.json field) to 3.6 to fall back to the old type definitions of `immer 5.02` (which I believe is the last version of  `immer` that was compatible to pre-3.7 TS versions).

I've built & packed this locally and un-tared it by hand into another project and tested it with TS 3.6.3. Without the entry in the `package.json`, my produced draft evaluated to `any`, with it it evaluated to the correct type, so this should be working.


markerikson:
@mweststrate : think you could get this into the next v6 alpha?

mweststrate:
Changed target to v6. TODO: add the new v6 api's, will see if I find some time later today

markerikson:
Sweet, thanks!

PR #549, opened 2020-03-04
rename compat `index.d.ts` to `immer.d.ts`

phryneas:
Heyo, sorry to bother you again :wave: 

Unfortunately, while I was opening #541, the field `types` in `package.json` was changed from `index.d.ts` to `immer.d.ts`, which again made the resolution of the compat types fail.
(See https://github.com/immerjs/immer/commit/ee1c977f5bf86e0f209e3df5f17bbc5bffb99dc3#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L12-R13 )

This should make it work again.


phryneas:
Just gave the CodeSandbox CI build of this PR a try over in redux-toolkit and it works nicely with 3.5 and 3.6: https://travis-ci.org/reduxjs/redux-toolkit/builds/658414463

mweststrate:
@phryneas no worries, should have totally caught this myself during merge. Thanks for the fix!

markerikson:
Thanks!  If you can toss out a 6.0.1 release with this, we'll use that right away in RTK.

mweststrate:
That part is automated, so it should hit npm in a couple of minutes :)

phryneas:
@mweststrate Um yeah, semantic release seems to be of the opinion that this is not worthy of a release :laughing: 
https://travis-ci.org/immerjs/immer/jobs/658441214

Could you trigger one manually? :)

mweststrate:
sigh

mweststrate:
yay! in some future I will be able to discern when the type: should be in
the first line, or a later line of the commit message...

Issue #550, opened 2020-03-12
404 error bit.ly link


techoi:
## Link to repro
[error.ts](https://github.com/immerjs/immer/blob/master/src/utils/errors.ts)

## To Reproduce

When error occur like ES5 on IE11
The error message is shown below
`[Immer] minified error nr: 19 ES5. Find the full error at: https://bit.ly/38PiBHb`

## Observed behavior
The url is not work.
404

## Expected behavior
Replace `https://bit.ly/38PiBHb` to `https://github.com/immerjs/immer/blob/master/src/utils/errors.ts`

## Environment

- **Immer version:**
- [ ] Occurs with `setUseProxies(true)`
- [x] Occurs with `setUseProxies(false)` (ES5 only)

mweststrate:
Yes, this is stupid, PR welcome!

techoi:
[PR](https://github.com/immerjs/immer/pull/551) fix this issue.

stunaz:
@techoi how did you fix this error? error 19 says the plugin is not loaded.... but I did load it as the following

```javascript
import {enableES5} from 'immer';

enableES5();
```

still getting the exact error

einasmadi:
@stunaz - I noticed that I needed to put the enableES5() line in my redux store initialization rather than at my Index.jsx entrypoint. I know it's been a while but if it helps anyone, if the error doesn't go it might be because the enabling is in the wrong place.

PR #551, opened 2020-03-13
fix: change Full error url (issue#550)

techoi:
As [issue#550](https://github.com/immerjs/immer/issues/550) said

[https://bit.ly/38PiBHb](https://bit.ly/38PiBHb) is not valid any more.

This PR just change url older to new one.

The new url origin url is `https://github.com/immerjs/immer/blob/master/src/utils/errors.ts`.
Shorten url is [https://bit.ly/3cXEKWf](https://bit.ly/3cXEKWf)

That's it.


mweststrate:
Thanks for doing this!

Issue #439, opened 2019-10-07
Calling an object getter on an immerable draft occurs an Error.


luncheon:
## 🐛 Bug Report

Calling an object getter on an `immerable` draft and mutate another property occurs an `Error`.

> TypeError: Cannot set property bar of #<Obj> which has only a getter

This error is not occured when `setUseProxies(false)`.

## To Reproduce

```javascript
import { produce, immerable } from "immer";

class Obj {
  constructor() {
    this.foo = 0;
    this._bar = { baz: 1 };
  }

  get bar() {
    return this._bar;
  }

  syncFoo() {
    return produce(this, state => {
      state.foo = state.bar.baz;
    });
  }
}

Obj.prototype[immerable] = true;

const obj = new Obj();
const obj2 = obj.syncFoo();
```

## Expected behavior

The error is not occurred.

## Link to repro (highly encouraged)

[CodeSandbox](https://codesandbox.io/s/damp-surf-lu81p)

## Environment

-   **Immer version:** 4.0.1
-   [x] Occurs with `setUseProxies(true)`
-   [ ] Occurs with `setUseProxies(false)` (ES5 only)

dalcib:
It can be fixe by the PR #438.

![image](https://user-images.githubusercontent.com/240801/66333934-96065480-e955-11e9-8cba-6791f8102147.png)

luncheon:
All right. Thank you. I'm looking forward to merging the PR!

dannyharding10:
Test comment

PR #558, opened 2020-03-23
Using a immerable class with a getter without a setter

dalcib:
In an immerable class with a getter without the corresponding setter, assigning a value to any field using the getter value within a class method causes the following error:

> #TypeError#
> Cannot set property bar of #<Obj> which has only a getter

It only happens with non-primitive getter values and with `setUseProxies(true)`
It can be fixed in userland creating an empty setter (`set prop(v) {}`) 
This PR makes the markChangedProxy function to skip the assignment `copy[key]=value` when there is a getter without a setter.

It fixes https://github.com/immerjs/immer/issues/439 and replaces https://github.com/immerjs/immer/pull/438

#### Links to Repro (updated to immer 6.0.2)
https://codesandbox.io/s/small-rain-inotu
https://codesandbox.io/s/useimmer-tic-tac-ci6qq

#### To Reproduce 
```js
import { produce, immerable } from "immer";
class Obj {
  [immerable] = true;
  foo = 0;
  _bar = { baz: 1 };
  get bar() {return this._bar}
  syncFoo() {
    return produce(this, state => {
      state.foo = state.bar.baz;
    });
  }
}
const obj = new Obj();
const obj2 = obj.syncFoo();
```

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -2090,6 +2090,46 @@ function testObjectTypes(produce) {
 			})
 		})
 	}
+
+	describe("class with getters", () => {
+		class State {
+			[immerable] = true
+			_bar = {baz: 1}
+			foo
+			get bar() {
+				return this._bar
+			}
+			syncFoo() {
+				this.foo = this.bar.baz
+			}
+		}
+		const state = new State()
+
+		it("should use a method to assing a field using a getter that return a non primitive object", () => {
+			const newState = produce(state, draft => {
+				draft.syncFoo()
+			})
+			expect(newState.foo).toEqual(1)
+		})
+	})
+
+	describe("class with setters", () => {
+		class State {
+			[immerable] = true
+			_bar = 0
+			set bar(x) {
+				this._bar = x
+			}
+		}
+		const state = new State()
+
+		it("should define a field with a setter", () => {
+			const newState3 = produce(state, d => {
+				d.bar = 1
+			})
+			expect(newState3._bar).toEqual(1)
+		})
+	})
 }
 
 function testLiteralTypes(produce) {
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -2108,8 +2108,11 @@ function testObjectTypes(produce) {
 		it("should use a method to assing a field using a getter that return a non primitive object", () => {
 			const newState = produce(state, draft => {
 				draft.syncFoo()
+				expect(draft.bar).toEqual({baz: 1})
 			})
 			expect(newState.foo).toEqual(1)
+			expect(newState.bar).toEqual({baz: 1})
+			expect(state.bar).toEqual({baz: 1})
 		})
 	})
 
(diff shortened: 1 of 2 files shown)


__tests__/base.js
```diff
@@ -2090,6 +2090,46 @@ function testObjectTypes(produce) {
...
+			}
+			syncFoo() {
+				this.foo = this.bar.baz
+			}
+		}
+		const state = new State()
+
+		it("should use a method to assing a field using a getter that return a non primitive object", () => {
+			const newState = produce(state, draft => {
+				draft.syncFoo()
```
mweststrate: could you assert the values of `draft.bar`, `state.bar` and `newState.bar` to be correct as well?

__tests__/base.js
```diff
@@ -2090,6 +2090,46 @@ function testObjectTypes(produce) {
...
+			set bar(x) {
+				this._bar = x
+			}
+		}
+		const state = new State()
+
+		it("should define a field with a setter", () => {
+			const newState3 = produce(state, d => {
+				d.bar = 1
+			})
```
mweststrate: could you assert `d._bar` and the original state's `_bar` as well after the test?

src/core/proxy.ts
```diff
@@ -239,8 +239,14 @@ export function markChangedProxy(state: ImmerState) {
...
 			each(state.drafts_!, (key, value) => {
-				// @ts-ignore
-				copy[key] = value
+				const descProto = Object.getOwnPropertyDescriptor(
+					Object.getPrototypeOf(copy),
+					key
+				)
+				if (!(descProto && descProto.get && !descProto.set)) {
+					// @ts-ignore
+					copy[key] = value
```
mweststrate: Q: If it is a getter, do we actually need to do anything at all or could we ignore the entire field?🤔 Since we should already pick up the change in the backing properties?

dalcib: Yes, that is the idea, just skip the field if it has a getter without a setter.

Another approach is to create an empty setter for these cases when creating the shallowCopy:

```js
src\utils\common.ts

export function shallowCopy(base: any, invokeGetters = false) {
	if (Array.isArray(base)) return base.slice()
+	const proto = Object.getPrototypeOf(base)
+	if (proto) {
+		Object.entries(Object.getOwnPropertyDescriptors(proto)).forEach(
+			([key, desc]) => {
+				if (desc.get && !desc.set) {
+					Object.defineProperty(proto, key, {
+						...desc,
+						set: x => {}
+					})
+				}
+			}
+		)
+	}
	const clone = Object.create(Object.getPrototypeOf(base))
```
I have tested and it also works well.

mweststrate: I misread something in the original, I think that one is fine as is :)

mweststrate:
Merged into #593, changed the implementation there to be a bit more efficient, and fix the problem and trap rather than at finalize time, still wip, but seems we can solve some other issues with it as well

PR #568, opened 2020-04-01
Ensure values in patches are never a draft

### LINKED ISSUE #559: A patch value field can sometime be a proxy
## 🐛 Bug Report

I am using patches to implement undo/redo, and `applyPatches` fails because a patch's `value` field is a revoked proxy. It looks like this happens because I am nesting reducers and my reducer's state is already a draft.

## Link to repro

Codesandbox repro: https://codesandbox.io/s/immer-sandbox-d1vg0

## To Reproduce

Open the repro, check in your browsers's console

## Observed behavior

The inverse patch's value is a Proxy (displayed as `null` by CodeSandbox's console, but correctly as a Proxy in your browser's console.

`applyPatches` fails to apply it, because the proxy has been revoked when applying the patch.

## Expected behavior

It applies the patch :)

## Environment

- **Immer version:** 5.3.6 & 6.0.2
- [ ] Occurs with `setUseProxies(false)` (ES5 only) => does not occur, as we dont have proper proxy revocation

renchap:
When you are generating patches in a nested reducer, patches values can be a Proxy. They need to be cloned before stored into the patch otherwise the patch cannot be applied as the proxy will be revoked.

I added a single test for this and ensured it was failing before my patch. If you want other tests to be added, please tell me what you would like.

Fixes #559

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -3,7 +3,8 @@ import produce, {
 	setUseProxies,
 	applyPatches,
 	produceWithPatches,
-	enableAllPlugins
+	enableAllPlugins,
+	isDraft
 } from "../src/immer"
 
 enableAllPlugins()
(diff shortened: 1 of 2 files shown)


renchap:
Hi @mweststrate! Any idea on when you will release a version with this fix?
Due to the build process I can not use my branch in my `package.json` to deploy this fix, and this bug is impacting a new feature we would like to launch soon.

Thanks!

mweststrate:
I'll expect tommorow😊

renchap:
Oops sorry for the `only`, good catch! And thanks for the release 👍

PR #595, opened 2020-05-06
Fix #594 Flow type for `Base` too limiting

### LINKED ISSUE #594: Flow type for `Base` too limiting
## 🐛 Bug Report

Some [recent changes to the flow types](https://github.com/immerjs/immer/commit/0dbcfe7205cb0ffdb56170da74fba61d8337d15d) seem to be too limiting: 

```js
type Base = {...} | Array<any> | Map<any> | Set<any>;
```

excludes `string | number | void | null` which all seem to be valid values to use with `immer`.

## Link to repro

Repository with reproduction case: https://github.com/esturcke/immer-flow-base-problem

## To Reproduce

```js
// @flow
import produce from "immer";

type Foo = { a: boolean } | void;

let foo: Foo = { a: false };
let bar = produce(foo, (draft) => {
  if (draft === undefined) return;
  draft.a = true;
});
```

## Observed behavior

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ index.js:7:11

Cannot call produce because undefined [1] is incompatible with function type [2].

     index.js
 [1]  4│ type Foo = { a: boolean } | void;
      5│
      6│ let foo: Foo = { a: false };
      7│ let bar = produce(foo, (draft) => {
      8│   if (draft == null) return;
      9│   draft.a = true;
     10│ });

     node_modules/immer/dist/index.js.flow
 [2] 33│         recipe: (draftState: S, a: A, b: B, c: C, ...extraArgs: any[]) => S | void,

Found 1 error
```
## Expected behavior

No flow errors.

## Environment

immer 6.0.5

nifgraup:
Allow currentState to be string, number & null as well as void (undefined) in the case of normal produce and curried produce with no initial state.

### DIFF
--- a/__tests__/flow/flow.js.flow
+++ b/__tests__/flow/flow.js.flow
@@ -100,6 +100,34 @@ type State = {| user: {| age: number |} |}
     }))
 }
 
+// #594 Allow undefined currentState with normal produce
+type Foo = {| a: boolean |} | void
+function func(foo: Foo) {
+    let bar = produce(foo, (draft) => {
+        // $ExpectError
+        draft.a = true
+        if (!draft) return
+        draft.a = true
+        // $ExpectError
+        draft.b = true
+    });
+    (bar: Foo)
+}
+
+// #594 Allow undefined currentState with curried produce and no initial state
+function func(foo: Foo) {
+    const f = produce(draft => {
+        // $ExpectError
+        draft.a = true
+        if (!draft) return
+        draft.a = true
+        // $ExpectError
+        draft.b = true
+    })
+    const bar = f(foo);
+    (bar: Foo)
+}
+
 {
     // $ExpectError
     produce2({x: 3})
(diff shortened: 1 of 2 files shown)


mweststrate:
See #603, in Immer 7 a non draftable base state will no longer be accepted as it is a source of confusion

nifgraup:
I'll drop this one then @mweststrate, but opened https://github.com/immerjs/immer/pull/625 with added tests for `Map` and `Set` as well as a simpler/fixed base type.

PR #597, opened 2020-05-09
docs: fix sample code

numb86:
It does not work as it now, and it is necessary to set the initial state.


numb86:
I have also set the initial state on the original reducer.

PR #625, opened 2020-06-18
Flow: Add tests for Map and Set and fix base type

nifgraup:
Map and Set are Objects in Flow so {...} covers them.


mweststrate:
Merged, thanks!

Issue #627, opened 2020-06-19
Map: new map key not added if value is undefined


visciang:
## 🐛 Bug Report

On Maps:
adding a **new** key with `undefined` value to a map does not create the new key.

## To Reproduce

Here a simple test to show the problem
https://github.com/immerjs/immer/compare/master...visciang:master

visciang:
Fixed in https://github.com/immerjs/immer/pull/630

PR #630, opened 2020-06-22
Map: new map key not added if value is undefined

visciang:
Fix for #627


mweststrate:
Awesome, thanks for the fix! Should be released soon.

PR #632, opened 2020-06-25
Flow: copy type of produce for produceWithPatches with adjustments inspired by the TypeScript types

prednaz:


mweststrate:
Thanks!

PR #634, opened 2020-06-28
Update performance.md - Fix a possible typo

jxjjligang:
I think there is a typo: skeeping should be skip.


prednaz:
There is "cool `Object.freeze(json)`" instead of "call `Object.freeze(json)`" too. When I noticed them before, I was not sure if they were not intended puns.

PR #653, opened 2020-08-13
skip ReadonlyMap and ReadonlySet types when not available

### LINKED ISSUE #624: TypeScript: ReadonlyMap & ReadonlySet not available when targeting `es5`
## 🐛 Bug Report

We had this as a bug report in redux-toolkit, but it boils down to immer.

If the `target` is set to `es5`, the `"ES2015.Collection"` library is not loaded by TypeScript, so `ReadonlyMap` and `ReadonlySet` are not available. This leads to the `Draft` type spinning into an endless recursion (as `ReadonlyMap` is interpreted as the empty object type, matching everything), resulting in a `Draft` type like
```ts
Map<Map<Map<Map<Map<Map<Map<Map<Map<Map<Map<..., ...>, Map<..., ...>>, Map<Map<..., ...>, Map<..., ...>>>, Map<Map<Map<..., ...>, Map<..., ...>>, Map<...>>>, Map<...>>, Map<...>>, Map<...>>, Map<...>>, Map<...>>, Map<...>>, Map<...>>
```
for the most simple drafted objects.

A simple fix would be to not rely on the external definitions of `ReadonlyMap` and `ReadonlySet`, but just to include them, which would mean defining

```ts
declare interface ReaddonlyMap<K, V> {
    forEach(callbackfn: (value: V, key: K, map: ReadonlyMap<K, V>) => void, thisArg?: any): void;
    get(key: K): V | undefined;
    has(key: K): boolean;
    readonly size: number;
}

declare interface ReadonlySet<T> {
    forEach(callbackfn: (value: T, value2: T, set: ReadonlySet<T>) => void, thisArg?: any): void;
    has(value: T): boolean;
    readonly size: number;
}
```

next to the `Draft` type.

If this would be a welcome addition, I'd be happy to open a PR :)

## Link to repro

A very simple reproduction can be found on https://github.com/jon-thompson/rtk-map-issue

## Environment

- **Immer version:**

phryneas:
This should fix #624 .

Instead of my initial idea of providing interfaces with the same shape as `ReadonlyMap`, `ReadonlySet`, `WeakMap` and `WeakSet` (which would also have worked, but probably required a lot more duplicate code), this is now following a different approach:

It checks if the types are `any` (which they would fall back to if they were not present) or a keyless interface (which at least `ReadonlyMap` and `ReadonlySet` would be if the `node` types are installed), in which case they fall back to `void`, which does not trigger the conditional.

The check for `any` is done by `true | false extends (T extends never ? true : false)` -  every other type will either be `true` or `false`, but since `any` takes both paths, it becomes `true | false`.

The check for the empty interface is done by `keyof T extends never` - everything except the empty interface has some keys, including primitives like `string` and `number` (which will have the keys of their prototype).

### DIFF
--- a/src/types/types-external.ts
+++ b/src/types/types-external.ts
@@ -18,16 +18,33 @@ type AtomicObject =
 	| String
 
 /**
+ * If the lib "ES2105.collections" is not included in tsconfig.json,
+ * types like ReadonlyArray, WeakMap etc. fall back to `any` (specified nowhere)
+ * or `{}` (from the node types), in both cases entering an infite recursion in
+ * pattern matching type mappings
+ * This type can be used to cast these types to `void` in these cases.
+ */
+export type IfAvailable<T, Fallback = void> =
+	// fallback if any
+	true | false extends (T extends never
+	? true
+	: false)
+		? Fallback // fallback if empty type
+		: keyof T extends never
+		? Fallback // original type
+		: T
+
+/**
  * These should also never be mapped but must be tested after regular Map and
  * Set
  */
-type WeakReferences = WeakMap<any, any> | WeakSet<any>
+type WeakReferences = IfAvailable<WeakMap<any, any>> | IfAvailable<WeakSet<any>>
 
 export type Draft<T> = T extends AtomicObject
 	? T
-	: T extends ReadonlyMap<infer K, infer V> // Map extends ReadonlyMap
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for the PR! Looks fine at first sight and clever trick :). However, if it turns out type checking becomes more expensive or there is other fallout, the changes might get referred.

PR #664, opened 2020-09-02
set assigned keys when clear map

### LINKED ISSUE #663: clear Map produces incorrect patches
## 🐛 Bug Report

clear Map produce incorrect patches

## Link to repro

codebox link: https://codesandbox.io/s/immer-sandbox-forked-zz8le

### immer source code
https://github.com/immerjs/immer/blob/bf90358c1bfedf0afb6e3bb016d6e6233b96e3ac/src/plugins/mapset.ts#L112
this line clears the `assigned_`, and Patch Plugin gen patches depends on `each(state.assigned_)`

https://github.com/immerjs/immer/blob/bf90358c1bfedf0afb6e3bb016d6e6233b96e3ac/src/plugins/patches.ts#L121-L143

is this by design? or a bug.

## To Reproduce

just run the  codesanbox link  above

## Observed behavior

`changes` is a empty array.

## Expected behavior

`changes` is `[{ type: 'Remove', path: ['t2', 't'] }]`

## Environment

- **Immer version:**

LYF1999:
fix #663

### DIFF
--- a/src/plugins/mapset.ts
+++ b/src/plugins/mapset.ts
@@ -15,7 +15,8 @@ import {
 	markChanged,
 	ProxyTypeMap,
 	ProxyTypeSet,
-	die
+	die,
+	each
 } from "../internal"
 
 export function enableMapSet() {
@@ -110,6 +111,9 @@ export function enableMapSet() {
 			prepareMapCopy(state)
 			markChanged(state)
 			state.assigned_ = new Map()
+			each(state.base_, key => {
+				state.assigned_!.set(key, false)
+			})
 			return state.copy_!.clear()
 		}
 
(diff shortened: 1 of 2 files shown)


src/plugins/mapset.ts
```diff
@@ -17,6 +17,7 @@ import {
 	ProxyTypeSet,
 	die
 } from "../internal"
+import {each} from "../utils/common"
```
mweststrate, line 20: NIT: import `each` from `internal` in the above import, as to guarantee there are no circular module deps

mweststrate (approved):
Looking good and thanks for adding a test as well! Left a small NIT

src/plugins/mapset.ts
LYF1999: thanks!, fixed it.

LYF1999:
hello, is there any other problem about this PR?

PR #669, opened 2020-09-11
Abuse typescript generics to show the error message for calls to `die`

forivall:
`doc: add generic to die to show what the error number means`

Mostly, I did this to help me understand the error codes quicker as i'm looking through how immer works.
 
Screen Shot 2020-09-11 at 4 01 45 PM
Screen Shot 2020-09-11 at 4 02 42 PM

Feel free to close if you don't want this in your codebase.

### DIFF
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -1,8 +1,9 @@
+type S = string & {}
 const errors = {
 	0: "Illegal state",
 	1: "Immer drafts cannot have computed properties",
 	2: "This object has been frozen and should not be mutated",
-	3(data: any) {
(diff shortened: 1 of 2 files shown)


src/utils/errors.ts
```diff
@@ -1,8 +1,9 @@
+type S = string & {}
```
forivall, line 1: In ts < 4, this prevents the const string literal from merging with the `string` type. can be removed to just "string" in ts 4.0+

mweststrate:
Closing as this makes the `die` definition itself now quite hard to read for just a few call sites :). I think a much simpler solution would be to add a JSDoc comment above those functions? Nonetheless thanks for sharing

PR #674, opened 2020-09-19
New tests (one currently failing): Non-immerable instances

bennorth:
I recently got tripped up by Immer silently misbehaving when the top-level `baseState` was a plain object, but within it was a class instance.  Had I read the manual more carefully, I would have been aware of the `immerable` requirement, but nonetheless it would have been helpful if Immer had thrown a noisy error.  This PR adds two tests, both testing that an error is thrown if you try to use Immer with an instance of a non-`immerable` class.  One of the tests fails, as no error is thrown for
```javascript
const Pet = class { constructor(name) { this.name = name } }
const baseState = {pet: new Pet("Rover")}
produce(baseState, draft => { draft.pet.name = "Rex" })
```
Would it be possible to have Immer throw its _[Immer] produce can only be called on things that are draftable..._ error in this case?  I'm afraid I'm not familiar enough with the internals of Immer to propose a fix, but hopefully these tests will be useful.

Thanks!

### DIFF
--- a/__tests__/__snapshots__/base.js.snap
+++ b/__tests__/__snapshots__/base.js.snap
@@ -22,6 +22,8 @@ exports[`base functionality - es5 (autofreeze) set drafts revokes sets 1`] = `"[
 
 exports[`base functionality - es5 (autofreeze) set drafts revokes sets 2`] = `"[Immer] Cannot use a proxy that has been revoked. Did you pass an object from inside an immer function to an async process? {}"`;
 
+exports[`base functionality - es5 (autofreeze) throws on non-immerable instance 1`] = `"[Immer] produce can only be called on things that are draftable: plain objects, arrays, Map, Set or classes that are marked with '[immerable]: true'. Got '[object Object]'"`;
+
 exports[`base functionality - es5 (autofreeze) throws when the draft is modified and another object is returned 1`] = `"[Immer] An immer producer returned a new value *and* modified its draft. Either return a new value *or* modify the draft."`;
 
 exports[`base functionality - es5 (autofreeze)(patch listener) async recipe function works with rejected promises 1`] = `"[Immer] Cannot use a proxy that has been revoked. Did you pass an object from inside an immer function to an async process? {\\"a\\":0,\\"b\\":1}"`;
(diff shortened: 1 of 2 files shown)


mweststrate:
The problem with this approach is that it will make it impossible to store arbitrary things in an immer tree, such as a Buffer, DomElement, Promise, etc etc.

bennorth:
I see what you mean.  As I say, I would not have been tripped up if I'd read the docs properly, but if there was a way to tell users of Immer that they're doing something unsupported, that would be helpful.  As a very rough suggestion, maybe in the example above,
```javascript
produce(baseState, draft => { draft.pet.name = "Rex" })
```
would throw an error, but
```javascript
produce(baseState, draft => {
  immerAccessNonDraftableProperties(draft).pet.name = "Rex"
})
```
would be OK?  Users can access or even mutate non-draftable properties, but only if they explicitly promise Immer that they know what they're doing.

The docs ([the "Classes" page](https://immerjs.github.io/immer/docs/complex-objects)) say _Immer does not support exotic objects such as DOM Nodes or Buffers_, so what should happen if someone has something non-draftable in their tree?  Currently, I think Immer lets you mutate your Buffer, and so the state is mutated, while keeping its identity.

I realise that the [introduction section](https://immerjs.github.io/immer/docs/introduction) of the docs can't explain every nuance, but currently it says _The interesting thing about Immer is that the baseState will be untouched_, which isn't the case in the above example.  Maybe an update to [the "Performance" page](https://immerjs.github.io/immer/docs/performance) would clarify things too; it currently says _Immer will convert anything you read in a draft recursively into a draft as well_, which I don't think is strictly always true.

(Would it be cleaner to close/reject this PR and move discussion to an issue, if you think there's a useful discussion to be had?)

mweststrate:
I think your solution still requires people to read the docs to avoid anything unexpected. Indeed this case isn't highlighted upfront in the docs, but storing classes instances in the first place is quite uncommon when the goal is to work with immutable data structures, and most tools, like for example redux middlewares, will require special handling if you do that.

bennorth:
OK, thanks for the discussion.  The main point of my suggestion was that an unexpected error would be better than an unexpected silent mutation of part of the state.  Then the user would know to go and read the documentation more carefully.

(For background: I am using Immer indirectly, via [easy-peasy](https://easy-peasy.now.sh/).  Its [quick-start documentation](https://easy-peasy.now.sh/docs/tutorials/quick-start.html#defining-actions-to-perform-state-updates) says that you define actions to "_Update the state by directly mutating the `state` argument.  Don't worry, under the hood we will convert the operation into an immutable update against your store by using Immer._"  I was then confused as to why my React component wasn't re-rendering after a change of state.)

mweststrate:
Ok, rewrote it as issue proposal #686, let me know if it covers what you are pointing at.

bennorth:
Thanks; your #686 captures the point well.

PR #677, opened 2020-09-23
Fix Clock time getter in complex objects docs

ivoelbert:
The time getter is trying to accessing `this.minutes`, but it's missing `this.`!
```ts
get time() {
	return `${this.hour}:${minute}`
}
```


mweststrate:
Thanks for noticing!

PR #682, opened 2020-10-09
Fix #680: Clearing empty Set&Map should be noop

### LINKED ISSUE #680: Clearing empty Set&Map should be noop
## To Reproduce

```
const { produce, enableMapSet } = require("immer");
enableMapSet();
const map = new Map();
const newMap = produce(map, draft => { draft.clear(); });
map === newMap
```
## Observed behavior

false
## Expected behavior

true
## Environment
- **Immer version:**

nifgraup:


nifgraup:
I was hoping the CI would run tests as they are red (before and after my change) when run locally.

mweststrate:
Looks CI succeeded and patch looks fine as well. Thanks!

PR #705, opened 2020-11-21
Use UTC date string in tests to be timezone independent

### LINKED ISSUE #712: Tests fail when run in non-UTC time zone
## 🐛 Bug Report

The current immer tests will fail if the system time zone is not UTC, e.g. when running locally instead of in CI. I built a fix for this and will PR it shortly. 

## Link to repro

n/a (it only reproes locally) 

## To Reproduce

```
npm test
```

## Observed behavior

```
 FAIL  __tests__/patch.js (5.203s)
  ● #676 patching Date objects

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `#676 patching Date objects 1`

    Snapshot: "2020-11-10T08:08:08.003Z"
    Received: "2020-11-10T16:08:08.003Z"

      1136 |    })
      1137 | 
    > 1138 |    expect(nextState.date.toJSON()).toMatchInlineSnapshot(
           |                                    ^
      1139 |            `"2020-11-10T08:08:08.003Z"`
      1140 |    )
      1141 |    expect(nextState.test.perform()).toBe("tested!")

      at Object.<anonymous> (__tests__/patch.js:1138:34)

  ● #676 patching Date objects

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `#676 patching Date objects 2`

    Snapshot: "2020-11-10T08:08:08.003Z"
    Received: "2020-11-10T16:08:08.003Z"

      1143 |    const rebuilt = applyPatches({}, patches)
      1144 |    expect(rebuilt.date).toBeInstanceOf(Date)
    > 1145 |    expect(rebuilt.date.toJSON()).toMatchInlineSnapshot(
           |                                  ^
      1146 |            `"2020-11-10T08:08:08.003Z"`
      1147 |    )
      1148 |    expect(rebuilt.date).toEqual(new Date(2020, 10, 10, 8, 8, 8, 3))

      at Object.<anonymous> (__tests__/patch.js:1145:32)

 › 2 snapshots failed.
```

## Expected behavior

No failures.

## Environment

- **Immer version:**
- [(n/a)] Occurs with `setUseProxies(true)`
- [(n/a)] Occurs with `setUseProxies(false)` (ES5 only)

Gelio:
I noticed that one of the tests failed when run locally (I am in the CET timezone ATM), because the stringified date did not match the snapshot (was 1 hour off):

image

This probably stems from the fact that the `Date` constructor parses the values as though they were in the local timezone (CET +1 for me). Going back to UTC (which is what `toJSON` produces) subtracts that 1 hour, which makes it show 7:08:08 instead of 8:08:08.

I reckon using a UTC string could solve the problem. Another alternative would be using a constructor with a UNIX timestamp, but I reckon the current solution is more readable.

Fixes #712

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -1131,7 +1131,7 @@ test("#676 patching Date objects", () => {
 	}
 
 	const [nextState, patches] = produceWithPatches({}, function(draft) {
-		draft.date = new Date(2020, 10, 10, 8, 8, 8, 3)
+		draft.date = new Date("2020-11-10T08:08:08.003Z")
 		draft.test = new Test()
 	})
 
@@ -1145,5 +1145,5 @@ test("#676 patching Date objects", () => {
 	expect(rebuilt.date.toJSON()).toMatchInlineSnapshot(
 		`"2020-11-10T08:08:08.003Z"`
 	)
-	expect(rebuilt.date).toEqual(new Date(2020, 10, 10, 8, 8, 8, 3))
+	expect(rebuilt.date).toEqual(new Date("2020-11-10T08:08:08.003Z"))
 })


mweststrate:
Thanks for fixing this!

PR #707, opened 2020-11-21
feature: Support copy-on-write in TypedArrays

### LINKED ISSUE #696: Typed Arrays support
## 🚀 Feature Proposal

Support typed arrays in Immer.

## Motivation

Sometimes typed arrays is just the best way to store data. One of the aspects is that if you use web workers, it is much more efficient to pass TypedArrays (Array Buffer) instead of copying things around. Also, from my understanding, those are just the most efficient data structure.

I guess I'm mostly curious whether that's actually possible and how much effort that would take.

Nope.

## Example
```
import produce from "immer"

const baseState = [
    {
        data: new Uint16(9000),
        data3D:[new Uint16(6756), new Uint16(789)]
    }
]

const nextState = produce(baseState, draftState => {
    draftState[0].data[93] = 0;
})

// or apply a patch
```

Gelio:
This is an attempt to fix #696 

This MR adds support for copy-on-write for TypedArrays (e.g. `Uint16Array`).

It correctly handles TypedArrays with the same underlying `ArrayBuffer`, so when copying 2 arrays with the same underlying buffer, the drafts will also have the same underlying buffer.

I was not sure how to handle ProxyTypes and Archtypes, so feel free to suggest another way to handle those. Also, do not hesitate listing things I could have done better 🙂 

## Limitations

I did not spend enough time to fix those. The known limitations/missing features are:

1. No support for patches
2. [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray](`TypedArray.subarray`) will not return a draft

## Possible optimizations

1. Not copying the array when the value is set to the same number

## Testing

Aside from the added automated tests, I did not test this feature extensively.

### DIFF
--- a/__tests__/__snapshots__/patch.js.snap
+++ b/__tests__/__snapshots__/patch.js.snap
@@ -5,3 +5,9 @@ exports[`applyPatches throws when \`op\` is not "add", "replace", nor "remove" 1
 exports[`applyPatches throws when \`path\` cannot be resolved 1`] = `"[Immer] Cannot apply patch, path doesn't resolve: a/b"`;
 
 exports[`applyPatches throws when \`path\` cannot be resolved 2`] = `"[Immer] Cannot apply patch, path doesn't resolve: a/b/c"`;
+
+exports[`typed arrays - patch throws for all \`op\`s 1`] = `"[Immer] TypedArrays cannot have patches."`;
+
+exports[`typed arrays - patch throws for all \`op\`s 2`] = `"[Immer] TypedArrays cannot have patches."`;
+
+exports[`typed arrays - patch throws for all \`op\`s 3`] = `"[Immer] TypedArrays cannot have patches."`;
--- a/__tests__/draft.ts
+++ b/__tests__/draft.ts
@@ -211,6 +211,13 @@ test("draft.ts", () => {
 		assert(toDraft(val), draft)
 	}
 
+	// Uint16Array
+	{
+		let val: Uint16Array = _
+		assert(toDraft(val), val)
+		assert(fromDraft(toDraft(val)), val)
+	}
+
 	// Promise object
 	{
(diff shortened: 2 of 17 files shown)


Gelio:
I have just noticed a mistake: `instanceof` and `ArrayBuffer.isView` will not work on the draft.

I'll add a test to ensure that those work and try to fix those

Gelio:
I have added a few more tests and commits, and changed the approach for what the proxy target should be in the case of typed arrays. I believe it should behave as expected

I'm looking forward to receiving feedback 🙂

mweststrate:
Sorry, currently not having a lot of bandwidth. Since this is quite a meaty change, probably will only review it in January or something, hope that isn't too much of a problem.

Gelio:
@mweststrate Sure, no problem :slightly_smiling_face: Take your time

mweststrate:
Closing as this feature never landed

Gelio:
I don't mind not having this feature in immer. I am wondering what do you mean by saying that this feature never landed. What feature never landed?

mweststrate:
This PR :)

Gelio:
Oh. It is disappointing to see a PR closed because it was not merged without any more justification. Especially that there was no delay and no asks from me at any point. I feel like my time was wasted. https://github.com/immerjs/immer/issues/696 is not closed and still encourages implementing this idea.

mweststrate:
Yeah sorry about that! Currently Immer has 5 million depend packages on Github, and the idea has 3 upvotes, so it seems hard to justify at this moment to ship the additional code and increased maintenance complexity to everyone. At this moment it'd become a classic example of feature creep I think :)

Gelio:
Sure, makes sense, thanks for a more thorough explanation :+1:

PR #706, opened 2020-11-21
chore: fix Buffer deprecation warning in test

Gelio:
There was a deprecation warning because `new Buffer([])` was used in one of the tests.

```
(node:25795) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
 PASS  __tests__/manual.js
```

image

PR #708, opened 2020-11-24
chore: use Buffer.from instead of deprecated new Buffer()

copperwall:
I noticed the test suite has a deprecation message about using `new Buffer([])` over `Buffer.from([])`. The buffer constructor usage was happening inside the test suite, so I thought I'd update the usage. The Node documentation says that using `Buffer.from` is equivalent, so I thought I'd update it. https://nodejs.org/docs/latest-v10.x/api/buffer.html#buffer_new_buffer_array

Running the test suite with this change removes the deprecation warning.


copperwall:
Oops, totally missed #706. I'll close this

Issue #686, opened 2020-10-21
Idea: forbid grabbing non draftables in strict mode


mweststrate:
## 🚀 Feature Proposal

See #674 and many other issues where people are surprised that Immer by default totally ignores their class instances. 
The idea is that if a field contains a non-draftable object, it becomes unreadable. E.g.: `produce(state, draft => { draft.buffer; })`  would throw. To still use the buffer inside a draft, a explicit signal needs to be given, e.g. `produce(state, draft => { mutableRef(draft.buffer).push(17); )`.

Not sure about the name: `mutableRef`, `unsafe` could all work. Or maybe we could even reuse the existing `original` method for this purpose, although that might be confusing. 

For easy backward compatibility an option `enableStrictMode(boolean)` could be used, by default off in the current version, and enabled in the next major. 

## Can this be solved in user-land code?

No

Gelio:
I tried to take a stab at implementing this in https://github.com/immerjs/immer/pull/715

PR #715, opened 2020-12-03
Throw when accessing non-draftables in strict mode

Gelio:
This MR attempts to implement the idea from #686

It adds a `strictMode` (enabled by `setStrictMode` or passing `strictMode` in `Immer` constructor) that makes it so that accessing non-draftable properties will throw an error, unless the access is made in an `unsafe` callback.

## Example

```ts
class Foo {}

produce({ instance: new Foo() }, draft => {
  unsafe(() => {
    draft.instance.abc = 1; // works fine in unsafe
  });

  console.log(draft.instance); // will throw, because access not wrapped in unsafe
});
```

See the tests for more examples

## Limitations

I tried to make it as compliant with existing functionalities as possible. I was not able to make it fully ES5-compliant. There are a few tests that fail when strict mode is enabled in `base.js` tests.

I do not have enough expertise to fully fix them. Suggestions are welcome

### DIFF
--- a/__tests__/strict-mode.js
+++ b/__tests__/strict-mode.js
@@ -1,0 +1,106 @@
+"use strict"
+import produce, {
+	setStrictMode,
+	unsafe,
+	immerable,
+	enableMapSet
+} from "../src/immer"
+
+enableMapSet()
+
+describe("Strict Mode", () => {
+	class Foo {}
+
+	describe("by default", () => {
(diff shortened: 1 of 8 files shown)


mweststrate:
Really awesome! Thanks for doing this :). Looks really promising, just up front a notif that it will probably take me a while to review 😅

Gelio:
@mweststrate I'm glad you like it! Take your time to review it

PR #727, opened 2020-12-26
fix: update comments

xyoscer:
Update comments to avoid misleading developers to troubleshoot bug


mweststrate:
Thanks for noticing!

PR #734, opened 2021-01-11
Update example-reducer.md

conradkirschner:
makes it more eaiser to see, where developers have to move their inital state. 

prevents some reading issues that creates stuff like: 
const byId = produce((draft = INITAL_STATE, action) => {


mweststrate:
Thanks!

PR #745, opened 2021-02-02
Gender neutral language in the docs

cpojer:
In the docs, I noticed there is an assumption in the example about the assistant's gender. This diff changes the text.

I had a version using "they" but when reading that back I was worried that non-native English speakers may get confused in this particular example since it is very early on in the docs and the example itself only serves as an elaboration of how draft state works. The version in this PR has the most clarity in my opinion, but I know that can be subjective. Happy to change it!


mweststrate:
I agree, in this case `the` works as well and is simpler. Thanks!

PR #754, opened 2021-02-18
Perf: Use true primitive types in AtomicObject

amcasey:
TypeScript will compare types to Boolean/Number/String structurally, but has special cases for boolean/number/string.  In local testing of a crazy edge case, this cut check time in half.


amcasey:
I bet I was supposed to fill out a PR template, but GH didn't present one when using the temp-fork workflow.  Please let me know if there's info you require.

PR #755, opened 2021-02-19
Add a type-checking fast path for primitive types

amcasey:
When `Draft<T>` is applied to a large enum type, TypeScript has to do a lot of unnecessary structural comparisons to confirm that no element of the enum matches `Function`, `Date`, `RegExp`, etc.  Determining that they do match `string` or `number`, on the other hand, is trivial.  This change splits `PrimitiveType` out of `AtomicObject` so that the fast path can be checked first.

In https://github.com/microsoft/TypeScript/issues/42824, this cut the check time from ~2.5 seconds to ~0.3 seconds.

### DIFF
--- a/compat/pre-3.7/dist/immer.d.ts
+++ b/compat/pre-3.7/dist/immer.d.ts
@@ -5,6 +5,8 @@ type Tail<T extends any[]> = ((...t: T) => any) extends (
 	? TT
 	: []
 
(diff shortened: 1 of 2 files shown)


amcasey:
Supersedes #754.

mweststrate:
Merged, thanks!

Issue #766, opened 2021-03-15
Not working when "produce()" object across iframes


atvoid:
When we have multiple iframes with the same origin, we can share objects between different iframe window. But immer produce has a check isPlainObject(), there is a condition "Object.getPrototype(xxx) === Object.prototype", which makes this scenario broken. Because the window.Object.prototype in different iframes have different instances. This condition will return false.

The basic repro example is:

in iframe, we can set window.parent.test = {a: 1}   (this can work when iframe and parent have same origin)
in parent window, we can access window.test and produce(window.test, (draft) => { } ), but this action will get exception "Not a plain object"

Can we change the condition "Object.getPrototype(xxx) === Object.prototype" in the function isPlainObject() ?

mweststrate:
Feel free to verify whether your solution or the one mentioned [here](https://github.com/immerjs/immer/issues/405#issuecomment-584549279) works and submit a PR :)

PR #767, opened 2021-03-18
fix isPlainObject to support cross frames

atvoid:
https://github.com/immerjs/immer/issues/766


mweststrate:
Thanks for making the PR!

PR #774, opened 2021-03-24
Update complex-objects.md

aschwartz-satcom:
Additional guidance on `[immerable]` when it comes to TypeScript and its structural typing system

### DIFF
--- a/website/docs/complex-objects.md
+++ b/website/docs/complex-objects.md
@@ -54,6 +54,44 @@ console.log(clock2.time) // 12:11
 console.log(clock2 instanceof Clock) // true
 ```
 
+### TypeScript and Structural Typing
+
+Type compatibility in TypeScript is based on structural subtyping. Structural typing is a way of relating types based solely on their members.
+For example:
+```js
+class Foo { bar: string }
+
+function logBar(someFoo: Foo) { console.log(someFoo.bar); }
+
+logBar({bar: 'Hello', someOtherProperty: 'ignore me'}); // OK in TypeScript
+```
+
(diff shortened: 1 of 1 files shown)


mweststrate:
Although it is true that class instances can be structurally the same as object literals, it is very dangerous to rely on that in practice (for example `instanceof` checks will fail!), and if you want to support such cases intentionally, I'd recommend to define an interface / type first. Since this is quite an edge cases and the TypeScript error given here is quite self explanatory, I think handling this cases explicitly is quite distracting in the docs. Nonetheless thanks for taking the effort to document it :) If you really love to see it in the docs, consider making it a one liner at the end of the typescript page instead, something along the lines: "Note: if you want to keep your immerable class typewise structurally equivalent to object literals, you can mark the `[immerable]` field as optional: `[immerable]? = true`.

aschwartz-satcom:
Thanks - I didn't realize there was a typescript-specific doc, I'll take a look at that one and change up my pr

On Mar 24, 2021 6:50 PM, Michel Weststrate ***@***.***> wrote:

Although it is true that class instances can be structurally the same as object literals, it is very dangerous to rely on that in practice (for example instanceof checks will fail!), and if you want to support such cases intentionally, I'd recommend to define an interface / type first. Since this is quite an edge cases and the TypeScript error given here is quite self explanatory, I think handling this cases explicitly is quite distracting in the docs. Nonetheless thanks for taking the effort to document it :) If you really love to see it in the docs, consider making it a one liner at the end of the typescript page instead, something along the lines: "Note: if you want to keep your immerable class typewise structurally equivalent to object literals, you can mark the [immerable] field as optional: [immerable]? = true.

mweststrate:
Closing for inactivity. Can be reopened if needed.

PR #800, opened 2021-05-20
Update built-with.md

chrisdugne:
added La Taverne https://github.com/uralys/taverne


mweststrate:
We try to limit the list to high profile projects, as otherwise the list will grow indefinitely :). Thanks for using Immer though!

Issue #804, opened 2021-05-26
Performance regression in `isPlainObject` utility 


benelliott:
## 🐛 Bug Report

The changes to  `isPlainObject` introduced in #767 added a more comprehensive check that worked for objects across iframes which relies on `Function.toString`:

```
	const Ctor =
		Object.hasOwnProperty.call(proto, "constructor") && proto.constructor
	return (
		typeof Ctor == "function" &&
		Function.toString.call(Ctor) === objectCtorString
	)
```

This more robust check is quite expensive for the common case where the object instance comes from the same frame and a simpler equality check would suffice.

I will submit a PR to remedy this.

## Link to repro

N/A (Perf)

## To Reproduce

N/A (Perf)

## Observed behavior

In our app we invoke `immer.applyPatches` multiple times per second based on WebSocket events. The calls to `Function.toString` have a noticeable performance impact on our app.

## Expected behavior

The utility method should not have a substantial impact on performance.

## Environment

We only accept bug reports against the latest Immer version.

- **Immer version:**
- [X] I filed this report against the _latest_ version of Immer
- [ ] Occurs with `setUseProxies(true)`
- [ ] Occurs with `setUseProxies(false)` (ES5 only)

PR #805, opened 2021-05-26
perf: isPlainObject: add quick comparison between input and `Object` to short-circuit taxing `Function.toString` invocations

benelliott:
Fixes https://github.com/immerjs/immer/issues/804

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -46,6 +46,10 @@ export function isPlainObject(value: any): boolean {
 	}
 	const Ctor =
 		Object.hasOwnProperty.call(proto, "constructor") && proto.constructor
+
+	if (Ctor === Object)
+		return true
+
 	return (
 		typeof Ctor == "function" &&
 		Function.toString.call(Ctor) === objectCtorString


mweststrate:
Thanks for optimising this!

markerikson:
FWIW, here's the implementation that Redux uses:

https://github.com/reduxjs/redux/blob/master/src/utils/isPlainObject.ts

```ts
/**
 * @param obj The object to inspect.
 * @returns True if the argument appears to be a plain object.
 */
export default function isPlainObject(obj: any): boolean {
  if (typeof obj !== 'object' || obj === null) return false

  let proto = obj
  while (Object.getPrototypeOf(proto) !== null) {
    proto = Object.getPrototypeOf(proto)
  }

  return Object.getPrototypeOf(obj) === proto
}
```

I believe this also correctly handles cross-frame behavior.  If I remember right, @timdorr implemented this.  Backstory:

- https://github.com/reduxjs/redux/issues/2598
- https://github.com/reduxjs/redux/pull/2599

PR #812, opened 2021-06-15
reduce size by ts const enum

houfeng0923:
hi, @mweststrate , thank you for great tool.  i review the code and got a  commit https://github.com/immerjs/immer/commit/fdc78db0ba08a68899722b2d32261207205bad95 which just for reduce bundle size . yes?

And [typescript const enum](https://www.typescriptlang.org/play?ts=3.9.7#code/MYewdgzgLgBApmArgWxgBQE4gB4E8AquADnDAN4BQAkJjrgPIBGAVnMLALwwCcANNbTwBBDBgCGuflQCiAZQCsTVuxhcADFLnyR4ydQCyYolNlwoFAL4UKAGzMwIq9FjyESAOlNQA3EA) may be  a better way for code and reduce ?

### DIFF
--- a/src/core/current.ts
+++ b/src/core/current.ts
@@ -8,8 +8,7 @@ import {
 	set,
 	ImmerState,
 	isDraftable,
-	ArchtypeMap,
-	ArchtypeSet,
+	Archtype,
 	getArchtype,
 	getPlugin
 } from "../internal"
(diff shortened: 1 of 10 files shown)


mweststrate:
@houfeng0923 thanks, this is great! Yeah const enums are definitely the better way to approach this :)

PR #821, opened 2021-06-27
fix typo from "nor" to "not"

anpleenko:


mweststrate:
The nor is correct here, as it should do at least one of them.

PR #833, opened 2021-07-30
doc: update example-setstate.mdx

VladislavMurashchenko:
Moved id generation for new todos out of reducers to keep them pure.

The main behind this PR is, that reducers should be pure functions. And it would be great if docs of other libs also don't violate this rule in their examples.
image

### DIFF
--- a/website/docs/example-setstate.mdx
+++ b/website/docs/example-setstate.mdx
@@ -140,7 +140,7 @@ const TodoList = () => {
 					break
 				case "add":
 					draft.push({
-						id: "todo_" + Math.random(),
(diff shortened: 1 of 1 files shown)


VladislavMurashchenko:
Note: probably these 2 sandboxes also should be updated:
https://codesandbox.io/s/immer-usereducer-bqpzn?file=/src/index.js
https://codesandbox.io/s/useimmerreducer-sycpb?file=/src/index.js

mweststrate:
Thanks for noticing! Updated the sandboxes as well

PR #860, opened 2021-10-18
async produceWithPatches

cviejo:
Fixes `produce` and `produceWithPatches` behaving differently when passed an async recipe. `produceWithPatches` returns:

```javascript
[ Promise { <pending> }, undefined, undefined ]
```
instead of a promise of `[nextState, patches, inversePatches]`


mweststrate:
Thanks!

PR #872, opened 2021-11-18
Append to array when key is "-"

### LINKED ISSUE #871: Appending to array using patch inserts element at beginning of array instead of the end
## 🐛 Bug Report

When using the `applyPatches` feature of immer with a patch that should append an element to the end of an array, the element is instead inserted at the beginning of the array.

## Link to repro

Using the [example from the JSON Patch RFC](https://datatracker.ietf.org/doc/html/rfc6902/#appendix-A.16):

>    An example target JSON document:
> 
>    { "foo": ["bar"] }
> 
>    A JSON Patch document:
> 
>    [
>      { "op": "add", "path": "/foo/-", "value": ["abc", "def"] }
>    ]
> 
>    The resulting JSON document:
> 
>    { "foo": ["bar", ["abc", "def"]] }

Doing so with immer instead creates the state:

```
//  { "foo": [ ["abc", "def"], "bar"]}
```

See https://codesandbox.io/s/immer-patch-append-ordering-repro-xi4d6 for a running version.

## Expected behavior

The element should be at the end of the array, not the beginning.

## Environment

- **Immer version:** 9.0.6
- [ ] Occurs with `setUseProxies(true)` n/a
- [ ] Occurs with `setUseProxies(false)` (ES5 only) n/a

lourd:
Closes #871

It's actually amazing '-' was doing anything at all! Definitely a JS quirk that the first argument to `splice` is coerced to a number, effectively becoming 0 if the parameter can't be parsed.

I also dropped an unnecessary `any` cast, and added a check for avoiding an unnecessary copy of the patch array.

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -720,6 +720,22 @@ describe("arrays - delete", () => {
 	)
 })
 
+describe("arrays - append", () => {
+	test("appends to array when last part of path is '-'", () => {
+		const state = {
+			list: [1, 2, 3]
+		}
+		const patch = {
+			op: "add",
+			value: 4,
+			path: ["list", "-"]
+		}
+		expect(applyPatches(state, [patch])).toEqual({
+			list: [1, 2, 3, 4]
+		})
+	})
+})
+
 describe("sets - add - 1", () => {
 	runPatchTest(
 		new Set([1]),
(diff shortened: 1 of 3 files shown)


lourd:
Please me know if there's I can do to help push this along! Tests are passing for me locally on Node.js v12 and v14; looks the failures on v16 are unrelated.

PR #889, opened 2022-01-09
toString method override bug patch

ranjit-git:
bug report https://huntr.dev/bounties/b08e508e-fa30-4692-b5ae-0f9b67ab20c7


mweststrate:
On hold; I don't think an intentional assignment to the `toString` property of a specific object is a security issue or even incorrect. Undesirable and confusing at most. I don't see how it pollutes the prototype, but if does, please do include tests in the PR that proof this, we don't accept PRs without regression tests.

PR #902, opened 2022-02-09
docs: add chinese translation

childrentime:
I added i18n to the configuration file,then add Chinese document translations. I've tested the build folder on my computer using nginx and it works fine.


iChenLei:
non-dark mode



dark mode



The language dropdown item's color is bad for looking when you not using dark mode.

mweststrate:
Thanks so much for adding these translations!❤️❤️❤️

childrentime:
❤️🧡💛💚💙💜🖤ʕ๑•ɷ•๑ʔ❀

PR #912, opened 2022-02-26
fix: special case NaN in setter

Rayzhangzhang:
I suppose there is a bug in setter about ‘NaN’ special case.


mweststrate:
would you mind adding a unit test to demonstrate the problem solved by this change?

Rayzhangzhang:
> would you mind adding a unit test to demonstrate the problem solved by this change?

Thanks for replay,  the unit test has been added.

unional:
Hi, @Rayzhangzhang this PR cause 8 test failures. thx

Issue #916, opened 2022-03-11
An error occurs when nesting patches and changing the level of [tree structure data]


Alt-er:
## 🐛 Bug Report

An error occurs when nesting patches and changing the level of [tree structure data]

## Link to repro

[https://codesandbox.io/s/minimal-repro-forked-xvb93p](url)

## To Reproduce

1.remove a node
2.Add a new node where it was removed
3.Add previously deleted nodes to a subset of new nodes

## Observed behavior

An error occurs when using nextData 

## Expected behavior

should not go wrong

## Environment

immer verson :9.0.12

Use default configuration

oriSomething:
I think it's related to how `Proxy.revocable()` works (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/revocable). 
You can use `enableES5(true)`, and `setUseProxies(false)` as work around. But I guess the bug related to `revoke()` to be called too early (means it's need to be called after all drafts finished, but I don't think it will solve all this kind of these issues).

PR #917, opened 2022-03-17
[minor] fix:change tree structure level in nested patches

Alt-er:
do not freeze drafts used by nesting

[issue](https://github.com/immerjs/immer/issues/916)


mweststrate:
Looks reasonable! Could you include a unit test demonstrating the original problem, and demonstrating that in the end the data is still getting frozen?

PR #921, opened 2022-04-04
fix: Fix importing ES module from Node.js (#901)

### LINKED ISSUE #901: immer is not properly exported to be used in Node.js ESM modules
I am trying to use immer in the backend of a Node.js application. I am currently in the process of migrating my backend to [ESM modules](https://nodejs.org/api/esm.html), since some dependencies (in particular [node-fetch](https://github.com/node-fetch/node-fetch)) are starting to ship only ESM modules.

# Error description

When I try to import immer in an mjs module using `import produce from 'immer';`, `produce` will be an object instead of a function, with its `default` property being the produce function.

I can access the produce function by using `import { produce } from 'immer';` or by using `import immer from 'immer';` and then using `immer.produce()`. The problem is that the documentation uses `import produce from 'immer';`, so there are libraries using this whose code I am not in control of. This means that these libraries break when I migrate to ESM modules.

# Reason for the error

immer is bundled in several different formats, among them `cjs` and `esm`. The bundles are referenced in `package.json` in the following way (`index.js` being a wrapper that includes the `cjs` bundle):
```json
  "main": "dist/index.js",
  "module": "dist/immer.esm.js",
```

While the `module` property is probably supported by webpack and other bundlers, it does not seem to be supported by Node.js. Instead, Node.js uses the `exports` property to support different main files for different environments (see [here](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions). Since that is not defined in this case, Node.js requires the file from the `main` property, which is a CommonJS bundle.

Even forcing Node.js to use the ESM bundle by doing `import produce from 'immer/dist/immer.esm.js';` does not solve the problem. The problem is that Node.js interprets files as CommonJS unless they have a `.jsm` file extension or `"type": "module"` is defined in `package.json` (which then applies to all files in the package) (see [here](https://nodejs.org/api/packages.html#determining-module-system).

# Possible solution

Setting `"type": "module"` is probably not an option, since that will break the CommonJS files.

The only solution that I can think of is to ship the ESM bundle as an `.mjs` file, either by renaming the current one or by creating a copy. The file can then be referenced in `package.json` like this:
```json
  "exports": {
    "import": "./dist/immer.esm.mjs",
    "require": "./dist/index.js"
  },
```

# Workaround

I have not yet found a workaround how I can migrate my app to ESM without breaking dependencies that use immer. If I find one, I will update the issue.

cdauth:
This pull request fixes #901 by publishing the ES module with the `.mjs` file extension.

For now, the existing ESM bundle `dist/immer.esm.js` is simply copied to `dist/immer.esm.mjs` building. This makes sure that there are no breaking changes, just in case any library includes the old file manually. In the next major release of Immer, the old bundle can be removed. By then jaredpalmer/tsdx#1059 might be released, which I believe will adapt the bundle file extensions.

This pull request also adds an `exports` field to `package.json`, which will make sure that Node.js will import the ES module when run in ESM mode. The wildcard export `"./*": "./*"` will make sure that any sub paths can be continued to be imported (for example something like `import 'immer/src/immer.ts'`, which makes this a non-breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,12 +3,19 @@
   "version": "9.0.0-beta.1",
   "description": "Create your next immutable state by mutating the current one",
   "main": "dist/index.js",
-  "module": "dist/immer.esm.js",
+  "module": "dist/immer.esm.mjs",
+  "exports": {
+    ".": {
+      "import": "./dist/immer.esm.mjs",
+      "require": "./dist/index.js"
+    },
+    "./*": "./*"
+  },
   "umd:main": "dist/immer.umd.production.min.js",
   "unpkg": "dist/immer.umd.production.min.js",
   "jsdelivr": "dist/immer.umd.production.min.js",
-  "jsnext:main": "dist/immer.esm.js",
-  "react-native": "dist/immer.esm.js",
+  "jsnext:main": "dist/immer.esm.mjs",
+  "react-native": "dist/immer.esm.mjs",
   "source": "src/immer.ts",
   "types": "./dist/immer.d.ts",
   "typesVersions": {
@@ -32,7 +39,7 @@
     "watch": "jest --watch",
     "coverage": "jest --coverage",
     "coveralls": "jest --coverage && cat ./coverage/lcov.info | ./node_modules/.bin/coveralls && rm -rf ./coverage",
-    "build": "rimraf dist/ && tsdx build --name immer --format esm,cjs,umd && yarn build:flow",
+    "build": "rimraf dist/ && tsdx build --name immer --format esm,cjs,umd && cp dist/immer.esm.js dist/immer.esm.mjs && yarn build:flow",
     "build:flow": "cpx 'src/types/index.js.flow' dist -v",
     "publish-docs": "cd website && GIT_USER=mweststrate USE_SSH=true yarn docusaurus deploy",
     "start": "cd website && yarn start",


mweststrate:
Looking good! Would you mind removing the `*` thing? I don't think there is any case where that is needed :)

mweststrate:
Merging as is. Typically whenever I touch something ESM related, the skies come tumbling down, so brace yourselves. Should be available as 9.1.0 in ~30 minutes

cdauth:
Nice! I'm also not 100% sure whether nothing will break, since there are so many module bundlers that handle modules in a different way. I'm curious to find out whether any reports will come in.

MoreSaltMoreLemon:
@cdauth 
> Nice! I'm also not 100% sure whether nothing will break, since there are so many module bundlers that handle modules in a different way. I'm curious to find out whether any reports will come in.

Hello!

One such breakage is coming from Redux Tookit, which has an [unpinned usage of immer](https://github.com/reduxjs/redux-toolkit/blob/2c6fab0b691c029a0a48e3bfc9bd59c7322fa203/packages/toolkit/package.json#L102)

v9.0.13 includes this PR breaks with React-Native's Metro module resolution for this package:

```
error: Error: While trying to resolve module `immer` from file `/Users/<user>/<app>/mobileApp/node_modules/@reduxjs/toolkit/dist/redux-toolkit.cjs.production.min.js`, the package `/Users/<user>/<app>/mobileApp/node_modules/immer/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/Users/<user>/<app>/mobileApp/node_modules/immer/dist/immer.esm.mjs`. Indeed, none of these files exist:

  * /Users/<user>/<app>/mobileApp/node_modules/immer/dist/immer.esm.mjs(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx)
  * /Users/<user>/<app>/mobileApp/node_modules/immer/dist/immer.esm.mjs/index(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx)
    at DependencyGraph.resolveDependency (/Users/<user>/<app>/mobileApp/node_modules/metro/src/node-haste/DependencyGraph.js:376:17)
    at Object.resolve (/Users/<user>/<app>/mobileApp/node_modules/metro/src/lib/transformHelpers.js:271:42)
    at resolve (/Users/<user>/<app>/mobileApp/node_modules/metro/src/DeltaBundler/traverseDependencies.js:571:33)
    at dependencies.reduce (/Users/<user>/<app>/mobileApp/node_modules/metro/src/DeltaBundler/traverseDependencies.js:587:26)
    at Array.reduce (<anonymous>)
    at resolveDependencies (/Users/<user>/<app>/mobileApp/node_modules/metro/src/DeltaBundler/traverseDependencies.js:586:33)
    at /Users/<user>/<app>/mobileApp/node_modules/metro/src/DeltaBundler/traverseDependencies.js:275:33
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/Users/<user>/<app>/mobileApp/node_modules/metro/src/DeltaBundler/traverseDependencies.js:87:24)
    at _next (/Users/<user>/<app>/mobileApp/node_modules/metro/src/DeltaBundler/traverseDependencies.js:107:9)
```

Anyone running into this: use `npx npm-force-resolutions` and pin the dependency to `9.0.12` instead.

cdauth:
@MoreSaltMoreLemon I haven't tried the new version yet, but the fact that I was having trouble with redux-toolkit was the original reason why I opened this pull request. I have also opened one with redux-toolkit (reduxjs/redux-toolkit#2200), but it has not been merged yet.

The error you are seeing looks strange to me, since the file `immer.esm.mjs` should actually exist. This seems to be a bug in metro: facebook/metro#535. Some workarounds are described there. I would expect more and more packages to use the `.mjs`/`.cjs` extensions, so I would say this problem needs to be solved in metro.

MoreSaltMoreLemon:
@cdauth  
> The error you are seeing looks strange to me, since the file `immer.esm.mjs` should actually exist. This seems to be a bug in metro: [facebook/metro#535](https://github.com/facebook/metro/issues/535). Some workarounds are described there. I would expect more and more packages to use the `.mjs`/`.cjs` extensions, so I would say this problem needs to be solved in metro.

I continued looking into it after my response and tried to recreate in a fresh create-react-native-app, and didn't. 

It appears that the issue is simply working with an older version of metro which predates .mjs support. Even though the file definitely exists, it isn't able to read it, and consequently tries to append a file extension that it can read to the provided filename and fails.

So resolutions are temporarily lock the version and update react-native/metro to a newer version with .mjs support.

rschristian:
Beyond the fields that Node consumes (and therefore requires the `(cm)js` semantics), it's generally better to keep `.js` for back compat.

It's only going to be an issue in older build tools, but regardless, this change was a semver major. `"module"`, `"jsnext:main"`, and `"react-native"` should be rolled back.

f-roland:
So things did break, for immer users but also users of libraries relying on immer (reduxjs/toolkit for one)

Shouldn't this be reverted and included in a major change instead ?

PR #927, opened 2022-04-15
Fix link to `return` on `produce` doc page

pzich:
The current format was linking to https://immerjs.github.io/immer/produce/return , which does not exist.

*Disclaimer: I'm not actually sure how to build/host/test these docs myself, but I based this change off of the changes in https://github.com/immerjs/immer/commit/bc359167ec53759f981f503a16a19a5e9a810159*


pzich:
I just tested with the `😎 Deploy Preview` linked above and it seems to work, but the path structure is different there so I'm not sure if it'll behave the same in production.

mweststrate:
LGTM, thanks!

PR #933, opened 2022-05-05
fix: incorrect return type for async produceWithPatches

mylesj:
👋 Hello, it looks to be a long standing definition but this line seems to be incorrect. 

For reference: 

Screenshot 2022-05-05 at 01 50 10

### DIFF
--- a/src/types/types-external.ts
+++ b/src/types/types-external.ts
@@ -248,7 +248,7 @@ export interface IProduceWithPatches {
 		base: Base,
 		recipe: (draft: D) => Promise<ValidRecipeReturnType<Base>>,
 		listener?: PatchListener
-	): PatchesTuple<Promise<Base>>
+	): Promise<PatchesTuple<Base>>
 }
 
 // Fixes #507: bili doesn't export the types of this file if there is no actual source in it..


weihomechen:
Has async  produceWithPatches  been removed in v10 ？

I  can't use like this:
```
 produceWithPatches({a: 0}, async d => {
        await Promise.resolve()
        d.a = 1
    }).then(([nextState, patches, inversePathes]) => {
    }))
```

mylesj:
@weihomechen Looks like a yes,

 - https://github.com/immerjs/immer/issues/1015

>  _**Drop promise based producer support.** Immer supports returning a promise from a producer. I don't think I ever saw usage of that in the wild, and it is a bit of an oxymoron as it makes it easy to leak drafts. Since producers should be like enter-write-exit without happening anything in the mean time, and it complicates the code base unnecessarily, I'm proposing to simply drop it._

weihomechen:
> @weihomechen Looks like a yes,
> 
> * [Immer 10 proposals: Drop promises, default export, array.length patches, ES5 mode, and make getter / setter support opt-in #1015](https://github.com/immerjs/immer/issues/1015)
> 
> > _**Drop promise based producer support.** Immer supports returning a promise from a producer. I don't think I ever saw usage of that in the wild, and it is a bit of an oxymoron as it makes it easy to leak drafts. Since producers should be like enter-write-exit without happening anything in the mean time, and it complicates the code base unnecessarily, I'm proposing to simply drop it._

I have to say it's a  bad  'news' ...
but thanks anyway

PR #935, opened 2022-05-10
[minor] Add failing test for nested and chained produce calls

jpallen:


mweststrate:
Putting on hold for now, as the test looks legit but contrived, so some context on the original problem this is addressing would be needed to be able to figure this one out.

mweststrate:
Ah, this might relate to #917, will check on top of that

jpallen:
Thanks for looking at this. This test is capturing the bug I described in https://github.com/immerjs/immer/issues/936, which has a bit more context as to what conditions triggered this error.

mweststrate:
NOTE: will merge this diff, which will break main, but otherwise I have to cherrypick the test and the original attribution to @jpallen  would be lost :)

Issue #937, opened 2022-05-11
Possible breaking change in package exports in v9.0.13


markerikson:
## 🐛 Bug Report

We just had a user file a Redux Toolkit bug report saying that as of Immer v9.0.13, RTK fails to initialize correctly, with a classic `Object(...) is not a function` error message

https://github.com/reduxjs/redux-toolkit/issues/2315

![image](https://user-images.githubusercontent.com/39653866/167955710-00aef1ac-e2fc-474b-99fc-247b417faa58.png)

Looking at the changes in https://github.com/immerjs/immer/compare/v9.0.12...v9.0.13 , I see that there was a recent PR changing how Immer does its exports:

https://github.com/immerjs/immer/pull/921

This seems like a likely candidate for causing this kind of problem.

Note that this change was _not_ called out in the 9.0.13 release notes, and also seems like it's a much bigger change than a "patch"-level release.  (FWIW I've been advised that changes to `exports` in `package.json` probably justify being breaking majors.)

## Link to repro

Don't have a repro yet - the end user just reported this a few minutes ago.

## Observed behavior

It looks like Immer's exports have potentially changed, depending on the build/run environment.  (Don't know details on the end user's environment yet).

## Expected behavior

Immer's imports load cleanly.

## Environment

We only accept bug reports against the latest Immer version.

- **Immer version:**
- [x] I filed this report against the _latest_ version of Immer

Rest of user environment is unknown, have asked for more details.

hirbod:
Your bundler most likely does not understand `mjs`. Had similar issues with React-Native and had to add `mjs` to `sourceExts` to make `Metro` happy

markerikson:
FWIW, it isn't "my" bundler :)  I'm a Redux Toolkit maintainer (RTK relies on Immer), and one of our users reported the bug against RTK.

Tbh this is the sort of thing that scares _me_ about trying to change package export setup for RTK itself, and why I refuse to touch our setup until we're ready for a major.

hirbod:
So it's your users bundler :)
But I'm totally with you on that. I wouldn't just do something like that in a minor release either.

And here a quote by one of the collaborators (@mweststrate)
> Merging as is. Typically whenever I touch something ESM related, the skies come tumbling down, so brace yourselves. 
https://github.com/immerjs/immer/pull/921#issuecomment-1087709203

Well, guess he has been right :)

f-roland:
related to https://github.com/immerjs/immer/issues/938

[...]

PR #939, opened 2022-05-12
fix: Use .esm.js module for backwards compatibility with old build tools

cdauth:
Some people have reported trouble with the new release of immer, in particular in #937 and #921. The reason seems to be that older build tools are having trouble to understand `.mjs` files as JavaScript.

Since currently immer publishes both an `.esm.js` and `.esm.mjs` bundle (introduced in #921), with the whole purpose of being backwards compatible, in this pull request I'm reverting the `module` and other fields in `package.json` to using the `.esm.js` bundle, while only the `exports` field uses the `.esm.mjs` bundle. I'm hoping that this will achieve both compatibility with Node.js as well as ancient build tools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "9.0.0-beta.1",
   "description": "Create your next immutable state by mutating the current one",
   "main": "dist/index.js",
-  "module": "dist/immer.esm.mjs",
+  "module": "dist/immer.esm.js",
   "exports": {
     ".": {
       "import": "./dist/immer.esm.mjs",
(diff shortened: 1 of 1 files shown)


f-roland:
Thanks @cdauth 

Fwiw, the same PR you did with a breaking change flag would have been just fine. 

What's the process to have this merged and released ?

mweststrate:
Merging this PR blindly to hopefully fix the issue! (I'm so tired of ESM)

hirbod:
I can confirm it fixed it for react-native.

HarshitMadhav:
Thank you @cdauth @mweststrate and @f-roland

PR #941, opened 2022-05-18
[breaking change] implement and default to useStrictShallowCopy, omitting getters

### LINKED ISSUE #867: Immer takes a long time to update the data.
## 🐛 Bug Report

Immer takes a long time to update the data.

## Link to repro

image

## To Reproduce

Chrome console:

```js
console.time('create');
const dataSource = Object.fromEntries(
  [...Array(4000).keys()].map((key) => [key, { key, data: { value: key } }]),
);
console.timeEnd('create');

console.time('update');
const result = produce(dataSource, (draft) => {
  draft[1000].data.value = 100;
});

console.timeEnd('update');
```

## Observed behavior

I checked the source code, the problem is in `shallowCopy`。If you use `Object.assign` instead, there is no performance problem

Time-consuming code:
```js
...
const descriptors = getOwnPropertyDescriptors(base)
...
return Object.create(Object.getPrototypeOf(base), descriptors)
```

## Expected behavior

Time shouldn't be so long

## Environment

- **Immer version:**

hrsh7th:
## Description

Fixes #867

This PR improves the performance when updating the object that contains a large number of entries.

### Problem

The current immer takes a long time for creating new object when the draft object has large amount of keys.

From some of the investigation, I found the cause that is `getOwnPropertyDescriptors` ([here](https://github.com/immerjs/immer/blob/master/src/utils/common.ts#L167)) .

In my computer, the `ownKeys` and its iteration don't take a long time but `getOwnPropertyDescriptors` takes time near the 10ms.

I can understand that the `getOwnPropertyDescriptors` is needed. Because immer should copy object's descriptors such as getter/setter/non-enumerable keys.

But some of the users doesn't need that strictness, I think.

### Solution

~~Add the ability to opt-out of strict shallow copy implementation for performance reasons.~~

`edited`

Make the default behavior not strict shallow copy. And add the ability to opt-in to strict shallow copying.

```ts
import { useStrictShallowCopy } from 'immer';

useStrictShallowCopy(true); // opt-in strict shallow copy.
```

### Result

The following results show the benefit of this patch.

```
// Iteration count:
//   50
//
// Base object: 
//   Array(10000)
//   	.fill(0)
//   	.map((_, i) => [i, i])
// 
// Test case:
//   for (let i = 0; i < MAX; i++) {
//   	produce(baseState, draft => {
//   		draft[5000]++
//   	})
//   }

# large-obj - mutate large object

immer (proxy) - with setUseStrictShallowCopy: 354ms
immer (proxy) - without setUseStrictShallowCopy: 79ms
immer (es5) - with setUseStrictShallowCopy: 1228ms
immer (es5) - without setUseStrictShallowCopy: 655ms
```

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -952,6 +952,7 @@ function runBaseTest(
 					enumerable: false
 				})
 
+				// TODO: Should I change the implementation to make the behavior unambiguous?
 				const canReferNonEnumerableProperty = useProxies || useStrictShallowCopy
 				const nextState = produce(baseState, s => {
 					if (canReferNonEnumerableProperty) expect(s.foo).toBeTruthy()
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -952,7 +952,7 @@ function runBaseTest(
 					enumerable: false
 				})
 
-				// TODO: Should I change the implementation to make the behavior unambiguous?
+				// When using Proxy, even non-enumerable keys will be copied if it's changed.
 				const canReferNonEnumerableProperty = useProxies || useStrictShallowCopy
 				const nextState = produce(baseState, s => {
 					if (canReferNonEnumerableProperty) expect(s.foo).toBeTruthy()
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -946,7 +946,7 @@ function runBaseTest(
 					value: 1,
 					enumerable: false
 				})
-				// Non-enumerable primitive property that don't refer
+				// Non-enumerable primitive property that won't modified.
 				Object.defineProperty(baseState, "baz", {
 					value: 1,
 					enumerable: false
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -9,7 +9,6 @@ import {
 } from "../src/immer"
 import {each, shallowCopy, DRAFT_STATE} from "../src/internal"
 import deepFreeze from "deep-freeze"
-import cloneDeep from "lodash.clonedeep"
 import * as lodash from "lodash"
 
 jest.setTimeout(1000)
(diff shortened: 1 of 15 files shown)


mweststrate:
I didn't read the code yet, but could you include a rationale in the PR description, what are you updating and why is it faster, what does the API look like, in which cases will it break things, etc?

gregdingle:
@hrsh7th @mweststrate this looks like a great PR to give users (like me) meaningfully better performance with large objects (>1000 keys). 

Was it closed just for lack of activity?

mweststrate:
It was missing rationale IIRC when it was closed, but now I see it is present, so reopening :) It might take some time to process though due to my own personal circumstances. Apologies upfront!

mweststrate:
I think it makes sense to make this behavior the new default, ship it as separate major version and make the old behavior opt-in using a flag.

hrsh7th:
Thank you for your review.

Nice to hear that this PR is an acceptable change.

I noticed that the test is failing. I will fix it.

hrsh7th:
Could you approve the GitHub Actions?

In my environment, this branch is passed the all tests... I want to re-try GitHub Actions.

mweststrate:
done!

__tests__/base.js
```diff
@@ -923,16 +946,26 @@ function runBaseTest(name, useProxies, autoFreeze, useListener) {
...
+					if (useStrictShallowCopy) expect(isEnumerable(s, "foo")).toBeFalsy()
+					if (canReferNonEnumerableProperty) s.bar++
+					if (useStrictShallowCopy) expect(isEnumerable(s, "foo")).toBeFalsy()
+					if (canReferNonEnumerableProperty) s.foo.a++
+					if (useStrictShallowCopy) expect(isEnumerable(s, "foo")).toBeFalsy()
+				})
+				if (canReferNonEnumerableProperty) expect(nextState.foo).toBeTruthy()
+				if (useStrictShallowCopy)
+					expect(isEnumerable(nextState, "foo")).toBeFalsy()
+				if (useStrictShallowCopy) expect(nextState.baz).toBeTruthy()
```
hrsh7th, line 968: This test is ugly, but after some consideration, I think it should be the spec.

First, I thought that non-enumerable properties should not be visible from drafts when useStrictShallowCopy is false.

However, achieving this behavior requires checking whether the referenced property is non-enumerable, which introduces new overhead.

hrsh7th:
I'm not confident about the documentation part, but I think the code part is ready for review.

PR #958, opened 2022-07-11
docs: fix incorrect argument position reference

dadamssg:


mweststrate:
Thanks for the PR, but it is incorrect, it is actually about the `first` argument, see the transformed listing in the second example :). If it was a confusing reading experience, do feel free to restructure!

dadamssg:
oohhhh okay. yeah i was hung up on that first part trying to reconcile what i was reading with what was being showed in the snippet. I didn't even scroll down. :expressionless:

PR #963, opened 2022-07-31
Jimmer doc moved and kotlin support

babyfish-ct:
1. Jimmer doc URL moved.
2. Kotlin is support, Now both Java and Kotlin can write immer style code.
3. Give example code and output


childrentime (approved):

mweststrate (changes requested):
Sorry for the late follow up, would you mind adding only a link? Since the other languages aren't maintained in this repo, documenting their usage here is unnecessary complicated.

babyfish-ct:
Sure, no problem

PR #964, opened 2022-07-31
[breaking change] Do not replace array.length as patch when removing items

kshramt:
Patches generated by the current implementation do not follow the RFC method for removing an element from an array (https://www.rfc-editor.org/rfc/rfc6902#page-13) and cannot be used from other languages (e.g., https://pypi.org/project/jsonpatch/).

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -555,7 +555,7 @@ describe("arrays - splice middle", () => {
 		},
 		[
 			{op: "replace", path: ["x", 1], value: 3},
-			{op: "replace", path: ["x", "length"], value: 2}
+			{op: "remove", path: ["x", 2]}
 		]
 	)
 })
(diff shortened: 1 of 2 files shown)


kshramt:
https://github.com/immerjs/immer/issues/208#issuecomment-996064385 might be resolved by the PR.

mweststrate:
Thanks for this PR! Apologies for the slow follow up, I do intend to merge, but as a separate major version.

mweststrate (approved):

bever1337:
I just submitted an MR that is in conflict with this. I was hoping to leverage REMOVE patches to describe sparse arrays. I agree with the use of operations here, but I'm not sure how to describe both operations. Any guidance? Thank you!

mweststrate:
Branch has been merged manually into #1028, thanks!

PR #969, opened 2022-08-18
protect against undefined constructor

BrianHung:
Edge case where constructor is not defined on an object, such as `Object.create(null)`.


mweststrate:
That looks correct! Would you mind adding a unit test to protect against regressions?

mweststrate:
Merged! Thanks again, and apologies for the late follow up.

PR #975, opened 2022-09-02
remove unused arg3 in produceWithPatches

BrianHung:
Reading the source code, I came across an unused `arg3`. Seems like it initially threw an error if it was provided to guard against developers mistakenly passing in a patch listener, so basically unsupported from the beginning:

https://github.com/immerjs/immer/blame/ee1c977f5bf86e0f209e3df5f17bbc5bffb99dc3/src/immerClass.ts#L171
https://github.com/immerjs/immer/commit/86be737d8502aeb351296afe28ceb3e506a3a2ec#diff-3a3d78b97c1f27bfee63f1d6c1ca4e45d828fecfc6ed4ec7be92eea461639d6bR101

### DIFF
--- a/src/core/immerClass.ts
+++ b/src/core/immerClass.ts
@@ -128,17 +128,17 @@ export class Immer implements ProducersFns {
 	}
 
 	produceWithPatches: IProduceWithPatches = (
-		arg1: any,
-		arg2?: any,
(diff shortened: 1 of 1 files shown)


mweststrate:
Looking good, thanks for the rename as well!

PR #976, opened 2022-09-09
[minor] Preserve insertion order of Sets, fixes #819

### LINKED ISSUE #819: Set order is not always maintained when modifying with produce
## 🐛 Bug Report

When adding to a Set with `produce`, the insertion order is not maintained when the original Set contains a draftable object.

## Link to repro

See #820 for a failing test (copied below)

```js
it("maintains order when adding", () => {
  const objs = [
    {
      id: "a"
    },
    {
      id: "b"
    }
  ]

  const set = new Set([objs[0]])
  const newSet = produce(set, draft => {
    draft.add(objs[1])
  })

  expect(Array.from(newSet)).toEqual([objs[0], objs[1]])
})
```

## To Reproduce

Run the "maintains order when adding" test in the "set drafts" section of `__tests__/base.js`.

## Observed behavior

After adding a second object to the Set with `produce`, it is ordered before the original first object in the Set.

## Expected behavior

`{ id: 'b' }` should appear after `{ id: 'a' }` when iterating through `newSet`, matching the behavior of a normal mutation-based update.

## Debugging Observations

As mentioned above, I've only noticed this in the situation where the original Set (pre-`produce`) contains a draftable object. For example, this test passes:
```js
it("maintains order when adding", () => {
  const objs = [
    "a",
    {
      id: "b"
    }
  ]

  const set = new Set([objs[0]])
  const newSet = produce(set, draft => {
    draft.add(objs[1])
  })

  // passes
  expect(Array.from(newSet)).toEqual([objs[0], objs[1]])
})
```
but the opposite does not:
```js
it("maintains order when adding", () => {
  const objs = [
    {
      id: "a"
    },
    "b"
  ]

  const set = new Set([objs[0]])
  const newSet = produce(set, draft => {
    draft.add(objs[1])
  })

  // does not pass
  expect(Array.from(newSet)).toEqual([objs[0], objs[1]])
})
```

## Possible Root Cause

(take all of this with a grain of salt - my first time digging into the repo 😅 )

After a bit of digging, this function seems to be causing the order change during `finalize`:
https://github.com/immerjs/immer/blob/dc3f66cdea53fd5a8c814924bfafa9f6b53c9c62/src/utils/common.ts#L131-L138

`finalizeProperty` is called on both objects (Draft of `{ id: 'a' }` and vanilla object `{ id: 'b' }`). On the first call, the Draft is converted back to a vanilla object and the Set value is replaced:
https://github.com/immerjs/immer/blob/dc3f66cdea53fd5a8c814924bfafa9f6b53c9c62/src/core/finalize.ts#L130-L131

Since Set values can't be directly updated, `set` deletes the previous version (Draft of `{ id: 'a' }`) and adds the new version (vanilla object), ultimately breaking the original order.

## Environment

- **Immer version: v9.0.3**

This appears to have been introduced in v5.2.0 which contains a rewrite of the Map and Set implementations, though I've only been able to test on v5.2.1 due to #502

jdip:
As pointed out in #819 insertion order is not always preserved when adding to a Set, though according to MDN it should be.

We were already cloning and iterating over the cloned Set anyway in `finalize`.

If we take it a step further to clear the original Set after cloning and let `finalizeProperty` know it needs to re-add non-draftable children back to the target then we end up rebuilding the Set in it's original order.

We then no longer need to delete anything from the Set in `set`

### DIFF
--- a/__tests__/map-set.js
+++ b/__tests__/map-set.js
@@ -296,4 +296,38 @@ function runBaseTest(name, useProxies, autoFreeze, useListener) {
 			expect(mapType1).toBe(mapType2)
 		})
 	})
+	describe("set issues " + name, () => {
+		test("#819.A - maintains order when adding", () => {
+			const objs = [
+				"a",
+				{
+					id: "b"
+				}
+			]
+
+			const set = new Set([objs[0]])
+			const newSet = produce(set, draft => {
+				draft.add(objs[1])
+			})
+
+			// passes
+			expect(Array.from(newSet)).toEqual([objs[0], objs[1]])
+		})
+
+		test("#819.B - maintains order when adding", () => {
+			const objs = [
+				{
+					id: "a"
+				},
+				"b"
+			]
+
+			const set = new Set([objs[0]])
+			const newSet = produce(set, draft => {
+				draft.add(objs[1])
+			})
+
+			expect(Array.from(newSet)).toEqual([objs[0], objs[1]])
+		})
+	})
 }
--- a/src/core/finalize.ts
+++ b/src/core/finalize.ts
@@ -87,12 +87,19 @@ function finalize(rootScope: ImmerScope, value: any, path?: PatchPath) {
 				: state.copy_
 		// Finalize all children of the copy
 		// For sets we clone before iterating, otherwise we can get in endless loop due to modifying during iteration, see #628
+		// To preserve insertion order in all cases we then clear the set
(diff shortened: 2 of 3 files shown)


mweststrate (approved):

mweststrate:
Sorry for the late follow up! Changes look legit, will include in the next minor.

jdip:
> Sorry for the late follow up! Changes look legit, will include in the next minor.

No worries. Thanks for taking the time to review/include it!

PR #980, opened 2022-10-04
docs: Fix broken cdn link due to trailing fullwidth period

luo3house:
I found the CDN script link was broken at installation docs in language zh_CN, which has redundant words because of trailing fullwidth period symbol, insert space can fix it.


mweststrate:
Thanks!

PR #989, opened 2022-10-30
feat: add `isNothing()`

### LINKED ISSUE #988: add `IsNothing()` type guard
## 🚀 Feature Proposal

Add `IsNothing()` type guard that can exclude `Nothing` from a union type.

## Motivation

The following does not work:

```ts
let x: string | Nothing | undefined

if (x === nothing) {
  // x type should be `Nothing`
}
else {
  // x type should be string
}
```

Since the `Nothing` type is implemented as a class, the `x === nothing` will not remove the `Nothing` type from the control flow analysis.

Adding a `IsNothing()` type guard will solve this problem.

I can implement that easily in user-land, but this is a core concept of `immer` and IMO should live inside `immer`.

I can contribute to the project. Just want to know which file should I add the tests to.

unional:
fix #988 

I can add tests, but want to know which file should I add it to. Type guard is TS stuff so maybe I create a new file `__tests__/common.ts`?

### DIFF
--- a/src/immer.ts
+++ b/src/immer.ts
@@ -3,7 +3,8 @@ import {
 	IProduceWithPatches,
 	Immer,
 	Draft,
-	Immutable
+	Immutable,
+	NOTHING as nothing
 } from "./internal"
 
 export {
@@ -17,12 +18,13 @@ export {
 	current,
 	isDraft,
 	isDraftable,
-	NOTHING as nothing,
 	DRAFTABLE as immerable,
 	freeze,
 	Objectish,
 	StrictMode
 } from "./internal"
+
+export {nothing}
 
 const immer = new Immer()
 
@@ -124,3 +126,7 @@ export {Immer}
 export {enablePatches} from "./plugins/patches"
 export {enableMapSet} from "./plugins/mapset"
 export {enableArrayMethods} from "./plugins/arrayMethods"
+
+export function isNothing(value: unknown): value is typeof nothing {
+	return value === nothing
+}


unional:
Another thing is, should the `Nothing` type be exposed?

Right now I have to access it through `typeof nothing`

mweststrate (changes requested):
Sorry for the late response! Feature looks good time, as does the test proposal. Would you mind to add the test file indeed, testing both the type inference and runtime behavior of the utility? Thanks!

unional:
> Would you mind to add the test file indeed, testing both the type inference and runtime behavior of the utility?

Sure, do you mind I add `type-plus` to the dev dependency so that I can easily test the types?

unional:
Also, looking at the PR again, it turns out the type is actually not exported. Will need to fix that.

The circular dependency is also an issue. Maybe I can create a different PR to fix all circular dependency issue first, as that would lead to many problems.

mweststrate:
> Sure, do you mind I add type-plus to the dev dependency so that I can easily test the types?

We already use `spec.ts` (see e.g. `produce.ts`), or using `// @ts-expected-error` might suffice?

unional:
I'll check that out. Thx

mweststrate:
@unional  were you able to set up some tests?

unional:
Hi, yes I just get back to this.

I have tried the `spec.ts` and `// @ts-expected-error` and saw how you use it:

```ts
// @ts-expect-error
assert(value, _ as Nothing)

// vs
isType.equal<false, typeof value, Nothing>()
```

I do found one issue from this PR.
The function `isNothing()` depends on the `Nothing` type, but `immer` implement this mechanism using the a dummy class `Nothing`.

I looked up the code and found that it is actually needed by one of the public type already:

```ts
/** Converts `nothing` into `undefined` */
type FromNothing<T> = T extends Nothing ? undefined : T

/** The inferred return type of `produce` */
export type Produced<Base, Return> = Return extends void
	? Base
	: Return extends Promise<infer Result>
	? Promise<Result extends void ? Base : FromNothing<Result>>
	: FromNothing<Return>
```

`Produced` -> `FromNothing` -> `Nothing`.

So exposing the type should be needed in the first place and that may be an existing bug. But testing and proving that is an issue could be difficult.

However, exposing this `Nothing` class is not good because it is not used as a class but a type.
One way to fix this is replace the mechanism with a branded type, e.g. https://github.com/unional/type-plus/blob/main/ts/nominal-types/Brand.ts#L11

But that will need some investigation.

PR #991, opened 2022-11-02
docs: Translate other-lang.md to Chinese language

Kennytian:


Kennytian:
@smichelJW @jscheid @marcins

PR #1007, opened 2023-01-03
fix: circular deps

unional:
There are still some remains, which needs to move code.

Will do it in subsequent commits

### DIFF
--- a/src/core/current.ts
+++ b/src/core/current.ts
@@ -1,17 +1,16 @@
+import {Archtype, ImmerState} from "../types/types-internal"
 import {
-	die,
-	isDraft,
-	shallowCopy,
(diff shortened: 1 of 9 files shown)


mweststrate:
The `internal` import trick is deliberately to control circular deps better, even though it generates the warnings in the build process, it does actually make the import order predictable, which is a lot more maintainable than reimporting stuff from all over the place, so I recommend to keep the structure as is :)

See https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de for the underlying rationale

unional:
Thanks! Good to know

PR #990, opened 2022-10-30
[minor] fix: recursive issue with Draft and JSONTypes

unional:
fixes #839 

I have added test under `__tests__/types/type-externals.ts`.

There are several things I have observed:

- the default formatter setting in vscode is not set to prettier. I have changed that so I won't mess up your formatting).
- there are some test failure from the previous PR
- the project doesn't build in Windows: `yarn build -> 'cp' is not recognized as an internal or external command`
- there are some circular dependency issues:

```sh
Circular dependency: src\internal.ts -> src\types\types-internal.ts -> src\internal.ts
Circular dependency: src\plugins\patches.ts -> src\immer.ts -> src\plugins\all.ts -> src\plugins\patches.ts
Circular dependency: src\internal.ts -> src\types\types-internal.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\utils\common.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\utils\plugins.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\scope.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\finalize.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\proxy.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\immerClass.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\current.ts -> src\internal.ts
Circular dependency: src\plugins\patches.ts -> src\immer.ts -> src\plugins\patches.ts
Circular dependency: src\plugins\patches.ts -> src\immer.ts -> src\plugins\all.ts -> src\plugins\patches.ts
Circular dependency: src\internal.ts -> src\types\types-internal.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\utils\common.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\utils\plugins.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\scope.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\finalize.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\proxy.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\immerClass.ts -> src\internal.ts
Circular dependency: src\internal.ts -> src\core\current.ts -> src\internal.ts
Circular dependency: src\plugins\patches.ts -> src\immer.ts -> src\plugins\patches.ts
```

[...]

PR #1014, opened 2023-01-16
Fix draft type for mixed property value types

### LINKED ISSUE #1013: Draft type no longer handles deeply nested readonly object with mixed property type correctly
## 🐛 Bug Report

The fix in https://github.com/immerjs/immer/pull/990 does not seem to properly fix https://github.com/immerjs/immer/issues/839 and also breaks other scenarios. It should probably be reverted.

That PR "fixes" the issue of recursive types by bailing out of the recursion if the property's value type does not extend `object`. This "works" in the case of `JsonObject` because `JsonObject`'s property's value type is a union of objects and primitives, and so it doesn't recurse on the object's properties because the property's value type does not extend the object type.

This is problematic because it won't remove the `readonly` modifier from nested objects if they have mixed property value types (both in the case of the `JsonObject` type and in other cases).

## Link to repro

```tsx
import { produce } from 'immer';

interface TestReadonlyObject {
    // The property value type does not extend object, so it will not recursively make the object writable.
    readonly testObjectOrNull: { readonly testProperty: number } | null;
}

const input: TestReadonlyObject = { testObjectOrNull: null };

produce(input, (draft) => {
    if (draft.testObjectOrNull) {
        // Errors because `testProperty` is still readonly, even though it shouldn't be.
        draft.testObjectOrNull.testProperty = 5;
    }
});
```

[Playground Link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzmKEAmBXAxgUzgXzgDNUQ4ByUEHKcgbgCgHgA7GGogQ1zgBUcAzjABKOTmggsANgE8A8gCMAVjizwEDOFrgB6HXwAWeFBDA0YMuADdOUjHgtm4EwXBYR4OAB7sWaOBDKqjAANHACEHDA8ADuwFJSbh5wUKoYUALAVjiycCCcANYORgFBanAxUNGcClI4AHSa2qnikrnsQooqanJQAHIYCQBciCliEtKWHTAACqhmsDIjLBggCjQEcAA+boNSjPhMWJJCUSxgGDAj-EKirZNdwXAAvKPTjz39e8t7BIwMJkwuAAFKwLqE4MC0FBOEQYABKF4APkQTS0wCIkOhsJg9XeZRgvQGCURGm05N0+gAolBUBk4OssJwMAI8AADaZzUzmGRsqICcIweKJFoTWRhHDZFhwGAGCAYADmBii8AEcsGaBY5Hg60aFO02LheMEhIJRL2xqEXIWFhecAArIxyYd8PDGEA)

## To Reproduce

See above.

## Observed behavior

There is a TypeScript error.

## Expected behavior

There should be no TypeScript error.

## Environment

- **Immer version:** 9.0.18

Methuselah96:
Fixes #1013.

This PR reverts https://github.com/immerjs/immer/pull/990 and adds a test that would fail without these changes to demonstrate the issue.


mweststrate:
Thanks for the PR! Releasing

Methuselah96:
@mweststrate Can you release this when you get a chance? The latest version on npm still has the broken types.

mweststrate:
Apologies! the release should have gone out automatically

Methuselah96:
No problem, thanks for taking a look.

mweststrate:
The release process has been fixed, should be available as 9.0.19 now

PR #1018, opened 2023-02-05
fix: ensure type exports is first

HiChen404:
## Summary
According to the typescript documentation: `Entry-point` for TypeScript resolution must occur first.

>Reference:  
> https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing


mweststrate:
Weird! Thanks for noticing

Issue #1019, opened 2023-02-26
v9.0.11 introduces sparse array regression


bever1337:
## 🐛 Bug Report

Background: https://github.com/immerjs/immer/issues/879

#879 looks like an awesome fix -- thank you! -- but it has a side-effect of breaking sparse arrays created with `delete`.

I'm a comfortable immer use but relatively new to the repo. If someone has the time to coach, I would be glad to iterate on #879. Thanks for your time

## To Reproduce

```javascript
const produce10 = require("immer-v9.0.10").default;
const produce19 = require("immer-v9.0.19").default;

/**
 * @returns {number[]}
 */
const initialState = () => [];

[
  ["v9.0.10", produce10],
  ["v9.0.19", produce19],
].forEach(([version, produce]) => {
  console.log(`Testing ${version}`);

  console.log("Produces [<empty>, 42]");
  // Create sparse array with insertion
  const stateA = produce(initialState(), (draft) => {
    draft[1] = 42;
  });
  console.info(stateA);
  expect(stateA[0], undefined, "Panic!"); // actually empty, but we expect undefined
  expect(stateA[1], 42, "Panic!");

  let iterateForEach = 0;
  stateA.forEach(() => {
    iterateForEach += 1;
  });
  expect(iterateForEach, 1, "Array.forEach does not iterate empty elements");

  let iterateForLoop = 0;
  for (let i = 0; i < stateA.length; i++) {
    iterateForLoop += 1;
  }
  expect(iterateForLoop, 2, "Delete does not alter array length");

  console.log("Produces [<empty>, <empty>]");
  // Create sparse array with delete
  const stateB = produce(stateA, (draft) => {
    delete draft[1];
  });
  console.info(stateB);
  expect(stateB[0], undefined, "Panic!"); // actually empty, but we expect undefined
  expect(stateB[1], undefined, "Panic!"); // actually empty, but we expect undefined

  iterateForEach = 0;
  stateB.forEach(() => {
    iterateForEach += 1;
  });
  expect(iterateForEach, 0, "Array.forEach does not iterate empty elements");

  iterateForLoop = 0;
  for (let i = 0; i < stateB.length; i++) {
    iterateForLoop += 1;
  }
  expect(iterateForLoop, 2, "Delete does not alter array length");

  console.log(`${version} success`);
});

/**
 * @param {boolean} condition
 * @param {string} reason
 * @returns {void}
 */
function expect(left, right, reason) {
  if (left !== right) {
[...]

PR #1020, opened 2023-03-01
fix: Creates separate patches for sparse arrays (#1019)

bever1337:
- the goal is to create a separate patch 'flow' for `delete` operations that cause arrays to become sparse
- es5 tests are currently failing

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -710,7 +710,7 @@ describe("arrays - splice (shrink)", () => {
 	)
 })
 
-describe("arrays - delete", () => {
+describe("arrays - delete property at index", () => {
 	runPatchTest(
 		{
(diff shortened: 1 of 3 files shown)


mweststrate:
Didn't read in detail, but JSONPatch (like JSON) doesn't seem to support a notion of sparse arrays, so any patch produced here would not behave consistent on non-immer clients: https://datatracker.ietf.org/doc/html/rfc6902/

bever1337:
Agreed these are not valid JSON patches. My thinking was that since only immer can reliably consume immer-made patches, there would be no harm in extending their internal functionality. This collides with the other PR though, which does treat array remove/replace according to spec, so I'm going to close this one. Thank you very much for your time! 😄

PR #1021, opened 2023-03-09
Fix typings for `union` types

isc30:
In cases where we have a union with `readonly` bits inside a `Draft`, the type is following the fallback to T instead of deconstructing the union (luckily for us, the union deconstruction is typed in the same way as object deconstruction, so the code can be simplified to simply fallback to `WritableDraft<T>` which covers both objects and unions).

This PR fixes very common use cases like `T | undefined`:

```ts
// BEFORE
type X = Draft<{ prop?: readonly string[] }>
//   ^? type X = { prop?: readonly string[] } // INCORRECT!

// AFTER
type X = Draft<{ prop?: readonly string[] }>
//   ^? type X = { prop?: string[] } // CORRECT!
```

### DIFF
--- a/src/types/types-external.ts
+++ b/src/types/types-external.ts
@@ -43,9 +43,7 @@ export type Draft<T> = T extends PrimitiveType
 	? Set<Draft<V>>
 	: T extends WeakReferences
 	? T
-	: T extends object
-	? WritableDraft<T>
-	: T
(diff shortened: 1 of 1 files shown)


isc30:
Closed as I noticed I wasn't using the latest version. Will check on the latest version and verify if its working well there

PR #1025, opened 2023-03-22
fix: patching maps failed when using number keys

### LINKED ISSUE #952: 'Cannot apply patch, path doesn't resolve' when modifying property of an object in Map
## 🐛 Bug Report

`applyPatches` throws error `Cannot apply patch, path doesn't resolve: map/1/b` when applying `inversePatches` generated by the following `produce` call:

```ts
interface Test {
  a: number;
  b: string;
}

let state = {
  map: new Map<number, Test>([])
};

state.map.set(1, { a: 5, b: "hello" }); // Setting initial state

let inversePatches: Patch[] = [];

state = produce(
    state,
    (draftState) => {
      const test = draftState.map.get(1);
      if (test) test.b = "world";
    },
    (_patches, _inversePatches) => {
      inversePatches = _inversePatches;
    }
);
```
It works fine when the patch op is `add` instead of `replace`:
```ts
state = produce(
    state,
    (draftState) => {
      const test = draftState.map.get(1);
      if (test) draftState.map.set(1, {...test, b: "world"}); // works
    },
    (_patches, _inversePatches) => {
      inversePatches = _inversePatches;
    }
);
```

## Link to repro
https://codesandbox.io/s/awesome-stitch-xxtxq3?file=/src/index.ts
- Click `Set Map Property` button to set map property
- Click `Undo` button to call `applyPatches` with inversePatches

## To Reproduce
- Call `enableMapSet` and `enablePatches`
- Create a `Map` with an object as the value type and add some key-value pairs to it
- Use `produce` to mutate the Map
- Inside `produce`, retrieve an element using `Map.get()` and modify a property on the element
- Save the `inversePatches` generated by `produce`
- Call `applyPatches(map, inversePatches)`

## Observed behavior
`applyPatches` throws error `Cannot apply patch, path doesn't resolve: map/1/b`.

## Expected behavior
The `inversePatches` should be applied successfully and `state` should revert back to the state it was in before the `produce` call

## Environment
- **Immer version: 9.0.14**

schummar:
This broke because of the mitigation for [CVE-2020-28477](https://github.com/advisories/GHSA-9qmh-276g-x5pj)

The solution should restore the behavior for maps with number key, while not bringing back the vulnerability.

Repro: https://stackblitz.com/edit/node-isqvrw?file=package.json
Fixes #952

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -214,6 +214,24 @@ describe("simple assignment - 7", () => {
 	)
 })
 
+describe("simple assignment - 8", () => {
+	runPatchTest(
+		new Map([[0, new Map([[1, 4]])]]),
+		d => {
+			d.get(0).set(1, 5)
+			d.get(0).set(2, 6)
+		},
+		[
+			{op: "replace", path: [0, 1], value: 5},
+			{op: "add", path: [0, 2], value: 6}
+		],
+		[
+			{op: "replace", path: [0, 1], value: 4},
+			{op: "remove", path: [0, 2]}
+		]
+	)
+})
+
 describe("delete 1", () => {
 	runPatchTest(
 		{x: {y: 4}},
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks for submitting this PR! Landing

Issue #1015, opened 2023-01-20
Immer 10 proposals: Drop promises, default export, array.length patches, ES5 mode, and make getter / setter support opt-in


mweststrate:
## 🚀 Feature Proposal

Hi folks, 

I've collected a bunch of issues that are together I think worthy of being put in a separate major version:

* [x] **Disable support for getters / setters by default**. Support getters and setters on plain object supports a non-idiomatic pattern where objects aren't purely data and can have side effects. Even though in general Immer does handle does correctly (descriptors are copied etc) there is are still 2 issues with it 1) it has a huge performance impact, and 2) side effects might be triggered non-intentionally. Given that getters / setters is an unidiomatic usage of "immutable" data, I'm proposing to make their support opt in. #1012 #867 #941 
* [x] **Drop promise based producer support**. Immer supports returning a promise from a producer. I don't think I ever saw usage of that in the wild, and it is a bit of an oxymoron as it makes it easy to leak drafts. Since producers should be like enter-write-exit without happening anything in the mean time, and it complicates the code base unnecessarily, I'm proposing to simply drop it. 
* [x] **Drop support for ES5 mode**. Proxies are now in all major browsers, and legacy support can stay on Immer 9, I think it is now the time to create a smaller (and simpler) Immer package by removing the ES5 fallback. 
* [x] **Drop the default export**. It is now valid to do both `import produce from 'immer'` and `import {produce} from 'immer'`. Since both default exports and offering two ways to achieve exactly the same are anti patterns, and removing simplifies ESM support, I'm proposing to drop the default export. This is quite a breaking change, but on the upside easily handled by a simple search and replace. https://github.com/immerjs/immer/issues/937#issuecomment-1386385087 / https://github.com/reduxjs/redux-toolkit/pull/3095
* [x] **Don't use array.length assignment in JSON patches**. See #964, array.length assignments should not be used to shorten arrays in JSON patches. Since this can both increase payload sizes and affect logic based on patches, saved this for a major bump.
* [x] **Give ESM another try**? See #921. Sigh. 
* ~~[ ] Support map/set by default -> avoid control issues like #1022~~

Notes to self:

* [x] TS 5
* [x] verify all error codes still in use
* [x] clean up old language constructions, iterators etc
* [ ] communicate that Immer now expects Symbol / Map / Set / Proxy to be present in env
* [ ] update size & perf tables
* [ ] chinese translation update
* [x] errors -> array

cc @markerikson

markerikson:
[...]

Issue #1012, opened 2023-01-10
Unexpected access to `getter` property in irrelevant plain objects


zthxxx:
## 🙋‍♂ Question

Here is two question, when will a plain object contain `drafts` in the recursively `finalize` function?

I think a plain object here always means not modify in `produce()` and not create by immer proxy, maybe move a proxy draft property into new object then add it back to draft's new property field?

<img width="806" alt="image" src="https://user-images.githubusercontent.com/15135943/211607547-c8c952b9-845b-435c-ad85-809babb0712f.png">

https://github.com/immerjs/immer/blob/8e12c78073c5f7c3a556a76dd780125b23bb12a3/src/core/finalize.ts#L62-L71

Is that the only thing need to do in this condition for plain object is `maybeFreeze(rootScope, value, true)` rather than `each(finalizeProperty(...)))` ?

I think this is the reason to cause some UNEXPECTED access to `getter` property in an irrelevant plain object when `produce()` returns result, which will have some side effects.

The problem case like below:

<img width="772" alt="image" src="https://user-images.githubusercontent.com/15135943/211635877-bd7ce9c8-13e6-4b2b-a69a-8ab843baa4be.png">

## Link to repro

To reproduce: https://stackblitz.com/edit/node-pgmcfz?file=immer-getter-without-freeze.mjs

## Environment

- **Immer version:** 9.0.17
- [ ] Occurs with `setUseProxies(true)`
- [ ] Occurs with `setUseProxies(false)` (ES5 only)

mweststrate:
Usually the quickest way to find these things out, is to disable that piece of code and see if a test breaks. Not sure if I understood your question correctly, but yes, I think it is indeed for "wrapping" cases like: `draft.x = { y: draft.y }`, which creates a plain object wrapping drafts. That is more common than you'd might expect since `draft.todos = draft.todos.filter(x => x.done)` does basically already create a plain object (well, array) wrapper around drafts.

zthxxx:
@mweststrate  This explanation is make scene to me, then I think the other reason in deep is the 93 line in `each` function,

it access `obj[key]` directly so that some `getter` will be called, but it's unnecessary in most of usage case call `each` function like "freeze property", on the other hand there are still a few cases need copy `getter` method or get value.

https://github.com/immerjs/immer/blob/8e12c78073c5f7c3a556a76dd780125b23bb12a3/src/utils/common.ts#L90-L98

[...]

PR #1028, opened 2023-03-23
Immer 10

### LINKED ISSUE #867: Immer takes a long time to update the data.

mweststrate:
# Release notes

* [breaking change] Immer 10 only supports modern browsers, that have support for `Proxy`, `Reflect`, `Symbol` and `Map` and `Set`.
* [breaking change] There is no longer a UMD build exposed (thanks [Mark Erikson](https://github.com/markerikson) for modernizing the build setup in #1032!
* [breaking change] getters and setters are ignored by default on plain object, as this is a very uncommon case and provides a significant performance boost (ca 33%, but depends a lot on the scenario). Fixes #867, #1012. Thanks [hrsh7th](https://github.com/hrsh7th) for implementing it in #941!
* [breaking change] Promise based reducers are no longer supported. Conceptually it is an anti pattern to hold on to drafts over time. If needed the old behavior can still be achieved by leveraging `createDraft` and `finishDraft`.
* [breaking change] ES5 mode (for legacy browsers) has been dropped. If your project relies on `enableES5()`, you SHOULD NOT upgrade Immer. `enableES5` has been removed.
* [breaking change] `produce` is no longer exposed as the `default` export. This improves eco system compatibility, and makes sure that there is only one correct way of doing things
* [breaking change] `enableAllPlugins` has been removed, use `enablePatches(); enableMapSet()` instead
* [breaking change] shortening the length of a JSON array now results in delete patches, rather than a mutation of the `length` property, in accordance with JSON spec. Thanks [kshramt](https://github.com/kshramt) for implementing this in #964!
* Immer is now an ESM package that can be directly imported into the browser. CJS should still work, UMD support has been removed.

Overall, there is a rough performance increase of 33% for Immer (and in some cases significantly higher), and a small overall package size reduction (TODO -> TODO bytes), with a small increase if only `produce` is imported: TODO -> TODO, or TODO -> TODO if everything is used. 

For more details, see #1015

## Migration steps

1. If you have any `enableES5()` call, don't migrate
1. When using getters/ setters icmw plain objects, call `useStrictShallowCopy(true)` at startup
1. Replace all default imports: Replace `import produce from "immer"` with `import {produce} from "immer"`
3. Replace all calls to `enableAllPlugins()` with `enablePatches(); enableMapSet();` to be more specific and smoothen future migrations. 
4. If any producer returned a Promise, refactor it to leverage `createDraft` instead. Roughly:

```javascript
const newState = await produce(oldState, recipe)

// becomes
const draft  = createDraft(oldState)
await recipe(draft)
const newState = finishDraft(draft)
```


__tests__/base.js
```diff
@@ -890,10 +896,16 @@ function runBaseTest(name, autoFreeze, useListener) {
...
-				foo: true
-			})
+			if (useStrictShallowCopy) {
+				expect(nextState).toEqual({
+					[test]: true,
+					foo: true
+				})
+			} else {
+				expect(nextState).toEqual({
+					foo: true
```
mweststrate, line 906: TODO: check why `[test]` is lost

src/core/current.ts
```diff
@@ -28,10 +28,14 @@ function currentImpl(value: any): any {
...
-		copy = copyHelper(value, archType)
+		copy = copyHelper(
+			value,
+			archType,
+			state.scope_.immer_.useStrictShallowCopy_
+		)
 		state.finalized_ = false
 	} else {
-		copy = copyHelper(value, archType)
+		copy = copyHelper(value, archType, true)
```
mweststrate, line 38: TODO: why `true`?

src/core/proxy.ts
```diff
@@ -17,6 +17,7 @@ import {
 	createProxy,
 	ArchType
 } from "../internal"
+import {ImmerScope} from "./scope"
```
mweststrate, line 20: TODO: import from internal

src/types/index.js.flow
```diff
@@ -83,6 +83,13 @@ declare export function setAutoFreeze(autoFreeze: boolean): void
  */
 declare export function setUseProxies(useProxies: boolean): void
```
mweststrate, line 84: TODO: remove

website/docs/api.md
```diff
@@ -30,6 +30,7 @@ title: API overview
 | `produceWithPatches` | Works the same as `produce`, but instead of just returning the produced object, it returns a tuple, consisting of `[result, patches, inversePatches]`. | [Patches](./patches.mdx) |
 | `setAutoFreeze` | Enables / disables automatic freezing of the trees produces. By default enabled. | [Freezing](./freezing.mdx) |
 | `setUseProxies` | Can be used to disable or force the use of `Proxy` objects. Useful when filing bug reports. |  |
```
mweststrate, line 32: TODO: remove

markerikson:
Hey, just saw this!  How's it looking so far?

mweststrate:
Mostly done except for docs / migration guide. I expect to release late
this week / next week

markerikson:
Nice!

Any details on perf or size changes?

markerikson:
Just ran a quick perf test with an RTK Query perf stress test scenario I've used previously.

I set it to render 1000 RTKQ-connected list items, then invalidated all the cache entries to force them to reload.

RTKQ dispatches when subscriptions are initiated, and again as data comes in.

With RTK 2.0.0-alpha.2 and Immer 9.x, average dispatch time during the load phase was 1.5ms.  With an RTK CSB CI build and Immer 10.0-beta.1, average dispatch time was 0.12ms (which is the lower limit of Chrome's function timing resolution):

image

When I hit "Invalidate", the Immer 9.x build had a solid 2.93 _seconds_ of dispatching.

With Immer 10, it was "only" 930ms:

image

The only change there was the Immer 10 bump.

So, clearly some pretty significant perf improvements!

That said, the Vite test project I'm using went _up_ in size with Immer 10.  The bundle went from 386K min (Immer 9.x) to 392K min (Immer 10).  Haven't looked into that yet, but very surprising.

mweststrate:
Those are awesome results! Better than in my (more artificial?) benchmark. Size increase is correct, because Map/Set usage is no longer opt-in but enabled by default. Let me know if you think that'd be a problem, but we had quite some issues related to people getting confuses on how to enable it for RTK (iirc) and other packages, as they accidentally imported immer a second time for example just to enable map/set, or already used map/set on the critical path, before the enabling method was called. So figured this is more frictionless, and most projects use Maps sooner or later anyway in practice

markerikson:
Just ran a quick check on 9.0.16 vs 10.0.beta, using Vite 4.2.1 + `source-map-explorer`:

- 9.0.16: `immer/dist/immer.esm.mjs`, 10.42KB
- 10.0.0-beta.4: `immer/dist/immer.mjs`, 15.78KB

If I hand-inspect the Vite bundle output, I don't _see_ any `NODE_ENV` references left in there, which is what I would expect - they exist in the ESM artifact, but the minifier should get rid of those.

If it helps at all, I just manually copied and pasted the minified versions of 9.0.16 and 10.0.0-beta.4 out of Vite-built bundles, ran them through Prettier, and am attaching them as a zip - maybe it'll help show where the differences are:


[immer-minified-prettified.zip](https://github.com/immerjs/immer/files/11140726/immer-minified-prettified.zip)



Per the `Map/Set` thing: FWIW, we really _don't_ want people using the `Map/Set` support with RTK because of serialization issues, so at least in our case that is a size increase we don't benefit from.  But, I do understand how that can cause confusion and annoyance on _your_ side.  (My own preference would be that it still be opt-in somehow if possible.)

mweststrate:
Ah interesting! Is there a documentation section I can refer people back to
if they report this in the future? RTK is probably the biggest packaged
Immer provider, so if that isn't even supposed to be done in that case,
happy to reconsider

markerikson:
Yeah, the two most immediate references are here:

- https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions
- https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

We actually had an internal discussion about `Map/Set` usage again yesterday after seeing your comment, and concluded it's still not sufficiently worth it.  It _is_ hypothetically possible to rework the Redux DevTools with one of the new "JSON superset" serialization libs, but it's fairly common for folks to want to persist state in `localStorage` (often using the `redux-persist` library), and that would run into serialization issues.  There'd also be more nuance to try to explain in the docs, and it's simpler if we leave it at a blanket "Don't!" :)

mweststrate:
@markerikson I reverted the MapSet opt-in in immer@10.0.0-beta.6, mind doing a check if it solves the bundle size increase?

markerikson:
@mweststrate yeah, I'll check on that tonight!

markerikson:
@mweststrate: Okay, 3 builds of the same RTKQ app, changing only the Immer version:

| Version     | Bundle (min) | Bundle (gzip) | Sourcemapped file |
|-------------|--------------|---------------|-------------------|
| 9.0.16      | 386.77 kB    | 122.41 kB     | 10.42 kB          |
| 10.0-beta.4 | 392.30 kB    | 123.74 kB     | 15.78 kB          |
| 10.0-beta.6 | 387.12 kB    | 122.31 kB     | 10.76 kB          |

markerikson:
@mweststrate : well, you got me hooked :)

I cloned the `immer-10` branch and did some poking around. I think there's a few things going on:

- You're currently building the project with TSDX.  That seems to be adding some polyfills, like `_extends()` (which you can see in the CJS dev build).
- That in turn is partly because `tsconfig.json` is set to `target: "ES6"`.  

So, those are adding up to a bunch of dead bytes that aren't needed.

Unrelated to the bundle sizes, there's also issues with the current build output in this beta:

- I know you've been flipping file names back and forth. Right now there's an `immer.mjs` bundle, but an `immer.esm.js.map` sourcemap file
- The sourcemaps themselves seem broken and incorrect.  If you try loading some of them into https://evanw.github.io/source-map-visualization/ , you'll see that variables and segments don't line up correctly, making them semi-useless

FWIW, I _just_ switched `redux` and `redux-thunk` over to use https://tsup.egoist.dev/ .  It's basically like TSDX conceptually, except that it's based on ESBuild + Rollup (and it seems to actually be _maintained_ atm).

I just tried making a local branch that switches over to `tsup` instead, and after 15-ish minutes I've got a decent part of the output bundles generating.  More than that, I've got the minified files shrinking noticeably.  In `beta.6`, `immer.cjs.production.min.js` is 16542b, and with `tsup` I've already got it down to 13680b.  I also tried generating a minified `.mjs` artifact for comparison, and that's down to 12941b.

Two other notes here:

- I see room for some byte shaving tricks.  For example, I tried `export const getPrototypeOf = Object.getPrototypeOf` and deleting the `Object.` in front of the uses, and that shaved off a few bytes because the variable name gets minified down to just `E`.
- I'm 95% set on dropping UMD bundles for Redux 5.0 / RTK 2.0, especially since you can make a pre-built ESM bundle that should be importable in-browser.  Might want to consider that here.

Sooooo, with all that said:

If you're interested, **I can try to do the rest of the TSDX -> `tsup` conversion, and file that as a PR against this branch**.

Thoughts?

mweststrate:
@markerikson thanks for pointing out tsup! I was already wondering what was the modern bundling setup since tsdx went unmaintained, which was also the reason I didn't want to poke too much in the config. So happy to receive a PR! Please do leave the flow types, as they're useful inside Meta, so I try to keep them roughly up to date :)

markerikson:
@mweststrate okay, sure!  I'll try to put up a PR over the weekend.

markerikson:
@mweststrate done! See #1032 .

Please let me know if you have any questions or would like me to tweak things further!

mweststrate:
Sorry had some PTO and now playing really hard catch-up! I'll try to go
over it in the weekend :)

mweststrate:
@markerikson just published `immer@10.0.0-beta.7`

PR #1029, opened 2023-03-29
chore(docs): use named imports for "produce" (instead of default import)

skratchdot:
I decided to update the docs after my comment here:
https://github.com/immerjs/immer/issues/1015#issuecomment-1487913925

All I did was change:
```
import produce from "immer"
```
to:
```
import {produce} from "immer"
```

The other formatting changes are due to the `pre-commit` hook:
image

I used `{produce}` instead of `{ produce }` to follow the other examples in this repo (even though I typically use the "extra spaces for padding").

Let me know if there's anything else I should do. Thanks for the great work on this project!

### DIFF
--- a/website/docs/async.mdx
+++ b/website/docs/async.mdx
@@ -35,7 +35,7 @@ sidebar_label: Async produce / createDraft
 It is allowed to return Promise objects from recipes. Or, in other words, to use `async / await`. This can be pretty useful for long running processes, that only produce the new object once the promise chain resolves. Note that `produce` itself (even in the curried form) will return a promise if the producer is async. Example:
 
 ```javascript
-import produce from "immer"
+import {produce} from "immer"
 
 const user = {
 	name: "michel",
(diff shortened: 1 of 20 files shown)


skratchdot:
what's odd is prettier seems to "flip-flop" the following html:

```
</center>
<details>
```
and:
```
</center> <details>
```

i use vscode with "auto format on save" and actually disabled it when making my changes.  but the `pre-commit` hook "re-added" the changes.

i'm just now noticing the diff, and the same "flip-flopping" happened.  Perhaps I should make my changes again, and commit with `-n` so the pre-commit hook doesn't run?

Or we could figure out how to prevent the format flip-flopping?  Perhaps a newline between the 2 tags would stop it? 🤷

skratchdot:
i'm updating the old commit (that used `pre-commit`):
cec1fa98b4743823299371f00a3fff94bb5f2e91

and using a new one (that skips the `pre-commit`):
0083ba3a4a3316d6a75065f88a64dca976061f9e

the diff looks a litter nicer now

mweststrate:
I suspect most of these changes are already made in #1028, you'd have to target branch immer-10 to be sure.

skratchdot:
@mweststrate - it looks like they are! i will close out this pr.  i still see some "flip-flopping" of the `</center> <details>` in that pr.  maybe i'll look into that at some point (but it doesn't seem like a big deal).

PR #1032, opened 2023-04-10
Switch build tooling to `tsup` and modernize JS build artifacts

markerikson:
This PR builds on top of the Immer 10 work in #1028 :

- Switches the build setup from `tsdx` to `tsup`
  - Updates the target JS output to be ES2020-ish (ie no transpiling at all)
  - Replaces the `__DEV__` flags with `process.env.NODE_ENV` checks
- Rearranges the build artifacts output to match what I've been setting up over in the various Redux repos for our next major releases. This combo appears to work well with all major build systems I've tested against, albeit with a `"FalseCJS"` warning from `arethetypeswrong`, which I'm ignoring for now
  - Adds an additional "legacy ESM" build that targets Webpack 4 for use with the `"module"` key, by using a `.js` extension and targeting ES2017 syntax (transpiling optional chaining and object spreads)
- Adds a micro-optimization by extracting `Object.getPrototypeOf` to a variable for better minification

This shaves a few KB off the size of a production build.  For the minified CJS Prod artifact:

- Immer 9.0.21: 16153b
- Immer 10.0.beta.6: 16535b
- This PR: 13680b

The minified production `.mjs` artifact is even smaller, at 12941b


mweststrate:
Awesome, thanks so much for introducing tsup, this looks so much better! Just merged the changes, with one minor change; I changed the `NODE_ENV==="development"` to `NODE_ENV!=="production"` to make development and test behave the same. I'm not sure whether there is a convention around this, if so, I can revert that and use different env config in the tests instead. 

Interestingly enough, after gzip etc the difference is especially reducing the MapSet plugin :) 

Anyway, thanks so much for this PR!

mweststrate:
Ah, just noticing that the private fields aren't minified yet, so the build can be a bit smaller still.

markerikson:
Awesome, glad this worked out!  

Let me know when you've got a new beta published!

markerikson:
@mweststrate One more possible trick I found.  Turns out `tsup` supports using Terser for minification instead of ESBuild.  I'm in the process of switching RTK over to use `tsup`, and here's how you configure it:

```ts
          minify: minify ? 'terser' : false,
          terserOptions: minify
            // usual Terser options here
            ? {
                output: {
                  comments: false,
                },
                compress: {
                  keep_infinity: true,
                  pure_getters: true,
                  passes: 10,
                },
                ecma: 2020,
                toplevel: true,
              }
            : undefined,
```

mweststrate:
I noticed that ESBuild also supports mangling props, do you know if terser
minifies better? Otherwise the simplicity of having a build tool less is
appealing, although terser never gave any issue so far

markerikson:
@mweststrate : Terser is a _bit_ better at minifying than ESBuild.

That said, I just tried copy-pasting your previous Terser settings over, and it was only like 50 bytes smaller _and_ the sourcemaps were borked.  So, sticking with ESBuild is probably the best option here.

benallfree:
@markerikson As others have mentioned in #1141, dropping UMD means Immer can no longer be loaded in the browser without going through a bundling tool. The installation instructions stating `<script src="https://unpkg.com/immer"></script>` are broken. I'm curious if not shipping a umd browser bundle was intentional?

markerikson:
@benallfree yes, dropping UMD builds was intentional - I did that for Immer here, and for RTK in 2.0.  Similarly, React 19 dropped UMD builds as well.

However, there _is_ now an `immer.production.mjs` file in the package that is specifically meant to be loaded as an ESM module in the browser, and that's intended to replace the need for UMD bundles:

- https://unpkg.com/browse/immer@10.1.1/dist/
- https://unpkg.com/immer@10.1.1/dist/immer.production.mjs

Probably true that I missed updating the README here to match that.

PR #1069, opened 2023-09-10
non-enumerable (immutable) methods on Map/Set

gimelg:
when an immer object includes a Map/Set, the mutating methods are replaced with ones that prevent mutation outside of the produce function.   to prevent cluttering the console when inspecting the Map/Set, these methods are set to non-enumerable. this behavior also matches that of the original methods that are replaced.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -188,7 +188,12 @@ export function freeze<T>(obj: T, deep?: boolean): T
 export function freeze<T>(obj: any, deep: boolean = false): T {
 	if (isFrozen(obj) || isDraft(obj) || !isDraftable(obj)) return obj
 	if (getArchtype(obj) > 1 /* Map or Set */) {
-		obj.set = obj.add = obj.clear = obj.delete = dontMutateFrozenCollections as any
+		 Object.defineProperties(obj, {
+                        set: {value: dontMutateFrozenCollections as any},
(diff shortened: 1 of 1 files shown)


mweststrate (approved):

src/utils/common.ts
```diff
@@ -188,7 +188,12 @@ export function freeze<T>(obj: T, deep?: boolean): T
 export function freeze<T>(obj: any, deep: boolean = false): T {
 	if (isFrozen(obj) || isDraft(obj) || !isDraftable(obj)) return obj
 	if (getArchtype(obj) > 1 /* Map or Set */) {
-		obj.set = obj.add = obj.clear = obj.delete = dontMutateFrozenCollections as any
+		 Object.defineProperties(obj, {
```
mweststrate, line 191: Looking good to me! As a nit optimisation, since the property map is constant, it could be lifted outside the function (usually that really doesn't matter, but with freeze potentially touching a gazillion of objects it might)

gimelg: Thanks! I'm waiting with sending the updated PR because the current state of the repo is failing some tests, both locally and using 'Open in Gitpod' (albeit a bit differently – locally __tests__/produce.ts, __tests__/draft.ts, and __tests__/redux.ts suits are failing, while in Gitpod the __tests__/immutable.ts is failing in addition to the other three).

Issue #1065, opened 2023-08-14
Unable to resolve "immer" from ....


jeyjeycodes:
## 🐛 Bug Report

I am building an Mobile App using React Native, Expo 49 (Hermes engine I believe).

I had installed the latest version of immer which is (10.0.2) My IDE is picking it up and I am able to use it perfectly fine. HOWEVER, when I try run it locally, I get this error:
![Screenshot 2023-08-14 at 15 22 10](https://github.com/immerjs/immer/assets/83839350/ce1c04f2-0214-4b5f-bd26-6c9cf01c8ffc)

So I then downgraded slowly and eventually when I hit version 9.0.21 it worked perfectly and was able to locally build and run the project smoothly :).

Also I am using yarn but I doubt that is the problem.

## Link to repro

My packages: 

```
{
	"name": "supabasetutorial",
	"version": "1.0.0",
	"main": "expo-router/entry",
	"scripts": {
		"start": "expo start --clear",
		"android": "expo start --android",
		"ios": "expo start --ios",
		"dev:client": "expo start --dev-client",
		"web": "expo start --web",
		"typecheck": "tsc --noEmit",
		"lint": "eslint '*/**/*.{js,ts,tsx}' --fix"
	},
	"dependencies": {
		"@expo/vector-icons": "^13.0.0",
		"@gorhom/bottom-sheet": "^4",
		"@gorhom/portal": "^1.0.14",
		"@hookform/resolvers": "^3.1.1",
		"@react-native-async-storage/async-storage": "1.18.2",
		"@react-native-community/datetimepicker": "7.2.0",
		"@react-navigation/material-top-tabs": "^6.6.3",
		"@react-navigation/native": "^6.1.7",
		"@supabase/supabase-js": "^2.2.0",
		"@tanstack/query-async-storage-persister": "^4.32.6",
		"@tanstack/react-query": "^4.29.19",
		"@tanstack/react-query-persist-client": "^4.32.6",
		"axios": "^1.4.0",
		"base64-arraybuffer": "^1.0.2",
		"dayjs": "^1.11.9",
		"expo": "^49.0.0",
		"expo-clipboard": "~4.3.1",
		"expo-constants": "~14.4.2",
		"expo-crypto": "~12.4.1",
		"expo-device": "~5.4.0",
		"expo-image-picker": "~14.3.2",
		"expo-linking": "~5.0.2",
		"expo-notifications": "~0.20.1",
		"expo-router": "2.0.0",
[...]

PR #1075, opened 2023-09-23
fix: don't use `.mjs` file for react-native, which isn't supported by default. Fixes #1058 #1065

### LINKED ISSUE #1058: While trying to resolve module `immer@10.0.2` React Native
Error: While trying to resolve module `immer` from file `/app/screens/comment.tsx`, the package `/build/node_modules/immer/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/node_modules/immer/dist/immer.mjs`. Indeed, none of these files exist

_Originally posted by @nhchuongchkst19 in https://github.com/immerjs/immer/issues/331#issuecomment-1654965627_

mweststrate:
pre-published as immer@10.0.3-beta

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "immer",
-  "version": "10.0.0-beta.6",
+  "version": "10.0.3-beta",
   "description": "Create your next immutable state by mutating the current one",
   "main": "./dist/cjs/index.js",
   "module": "./dist/immer.legacy-esm.js",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     }
   },
   "jsnext:main": "dist/immer.mjs",
-  "react-native": "dist/immer.mjs",
+  "react-native": "./dist/immer.legacy-esm.js",
   "source": "src/immer.ts",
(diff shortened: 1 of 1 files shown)


markerikson:
It looks like RN 0.72 will help with this by finally supporting `exports` properly, but yeah, this might be necessary to support the rest of the ecosystem until they catch up:

- https://reactnative.dev/blog/2023/06/21/package-exports-support

markerikson:
Looking at https://facebook.github.io/metro/docs/package-exports/ , it sounds like Metro currently looks for `"react-native"`, `"browser"`, and `"main"`:

- https://facebook.github.io/metro/docs/package-exports/
- https://github.com/facebook/metro/blob/179205b908b6ed4e458f35b852bb8304bb225652/docs/Configuration.md?plain=1#L230

I see that Immer actually does list `"react-native"`.  The Redux packages don't, but do list `"main"`, and in our current Redux betas, I _am_ pointing `"main"` to that similar `legacy-esm.js` file.  So, I _think_ we've got the same net result.

PR #1084, opened 2023-12-21
Update installation.mdx

MUHAMMADMUKHTARSULEIMAN:
It read "So an url," so I merely changed the an to a.


mweststrate:
I'm not a native speaker either, but afaik "an URL" is grammatically correct since the acronym starts with a vowel.

mweststrate:
Quick google says you're right though since the "U" is pronounced as a consonant. :) Ugh, English. Merging.

PR #1088, opened 2024-01-10
Add Optional Support For Multiple References to an Object

BellCubeDev:
Some state trees may need to reference an object more than once (such as the tree for my [fomod](https://www.npmjs.com/package/fomod) library). By using a combination of a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to store existing drafts and an off-by-default configuration option, this should be a painless solution. I've tested it within the scope of my project but there may always be issues I haven't foreseen.

The diff is rather small and I fully expect that there will be features and use cases of Immer that enabling this config option may break. This is my first time looking behind the curtain of Immer, after all.

### DIFF
--- a/src/core/immerClass.ts
+++ b/src/core/immerClass.ts
@@ -34,8 +34,13 @@ interface ProducersFns {
 export class Immer implements ProducersFns {
 	autoFreeze_: boolean = true
 	useStrictShallowCopy_: boolean = false
+	allowMultiRefs_: boolean = false
 
-	constructor(config?: {autoFreeze?: boolean; useStrictShallowCopy?: boolean}) {
+	constructor(config?: {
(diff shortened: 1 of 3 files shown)


src/core/immerClass.ts
```diff
@@ -194,20 +212,30 @@ export class Immer implements ProducersFns {
 			applyPatchesImpl(draft, patches)
 		)
 	}
+
+	existingProxyMap_: WeakMap<Objectish, Drafted> | null = null
```
BellCubeDev, line 216: Was testing the behavior in my own tool and realized that, when a draft is finalized, the two objects will be different objects rather than the same object. I'll need to add a `draftToStateMap_` as well.

Issue #1091, opened 2024-02-18
How to deep clone ignoring immer references? (deepCurrent vs currrent)


lveillard:
Hello! 

**Issue**
I'm struggling to mutate an object as i need to split it and then add different things to each new end. The problem is that the original object has nested objects, so when duplicating, the original nested objects get referenced.

Lets say i have
```
const a = {b: 0}
```
And then i split a to:
```
a => [a1, a2]
a1 => {b:0}
a2 => {b:0}
```

if i do this in the same draft and i modify b:2 it affects both.

I'm trying to sue structuredClone, structuredClone(current(a)) but they don't work because the draft are embeded.

With some real data:
```
subNode {
      '$thing': 'User',
      '$thingType': 'entity',
      '$op': 'update',
      '$bzId': 'R_9b59cc31-9df8-44ad-9dc2-839108f3f116',
      '$entity': 'User',
      '$id': <ref *1> [....} ],
      spaces: <ref *2> [
        {
          type_: 1,
          scope_: [Object],
          modified_: false,
          finalized_: false,
          assigned_: {},
          parent_: [Object],
          base_: [Array],
          draft_: [Circular *2],
          copy_: null,
          revoke_: [Function (anonymous)],
          isManual_: false
        }
      ]
    }
```
   Basically I split the original object which has $id: ['id1' ,'id2'] into an array with two objects.
 ```
 	return subNode.$id.map(($id: string, i: number) => ({
	   ...subNode,
	   $id: $id,
	   $bzId: `${subNode.$bzId}_${i}`,
	}));
```
This works at the root node, i get an array with two different BzId "_0" and "_1"
But if i modify the nested object, it gets modified in both places.

I tried to use this: `...structuredClone(subNode),`

But it gets this error:  `[object Array] could not be cloned.`
I guess this happens when it arrives to the nested structures created by immer.

Is there a way in immer to deeply clone an object and drop?

or the only way i can do right is a deepCurrent() and then use structureCloned over the result?

**Immer version:**
Using immer (9.0.21 as 10.0.x does not work with symbols that have references to objects)

mweststrate:
This might be a bug in `current`, but we need a repro to be sure. (See also comments in the PR)

mweststrate:
#1105 should fix the handling of Proxy properties. When drafted, they are now finalized correctly. Let me know if it isn't solved by `immer@10.0.4`

lveillard:
[...]

PR #1092, opened 2024-02-18
feat: deepCurrent

lveillard:
Related to this issue: https://github.com/immerjs/immer/issues/1091

This is just a draft, if it goes validated I can test it, add some tests and apply any changes you might ask for!

### DIFF
--- a/src/core/deepCurrent.ts
+++ b/src/core/deepCurrent.ts
@@ -1,0 +1,45 @@
+import {
+    die,
+    isDraft,
+    shallowCopy,
+    each,
+    DRAFT_STATE,
+    set,
+    ImmerState,
+    isDraftable,
+    isFrozen
+} from "../internal"
+
+// Extend the current function to handle deep drafts
+export function deepCurrent<T>(value: T): T
+export function deepCurrent(value: any): any {
+    if (!isDraft(value)) die(10, value) // Use Immer's error handling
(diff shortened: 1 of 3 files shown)


mweststrate:
Thanks for the effort of creating this PR! I think this might actually be a bug in the current `current` (lol) implementation, as it is supposed to be able to handle recursion already by the looks of it: https://github.com/immerjs/immer/blob/main/src/core/current.ts#L33

From a quick scan your implementation seems also to be exactly the same as `current`? Would you mind first setting up a unit test showing the problem in the existing `current` implementation?

lveillard:
> Thanks for the effort of creating this PR! I think this might actually be a bug in the current `current` (lol) implementation, as it is supposed to be able to handle recursion already by the looks of it: https://github.com/immerjs/immer/blob/main/src/core/current.ts#L33
> 
> From a quick scan your implementation seems also to be exactly the same as `current`? Would you mind first setting up a unit test showing the problem in the existing `current` implementation?

True! Somehow deepCurrent works for my use case and current does not 🤷‍♂️. Will try to find some time to build that test triggering the issue, I guess is related to using symbols with objects as values.

mweststrate:
Closing in favour of #1105, that fixes the underlying issue (I think). Thanks for reporting and taking the effort for this PR though!

PR #1094, opened 2024-02-27
Make  `applyPatches` to accept  `readonly Patch[]`

Andarist:
My main goal is for `applyPatches` to accept `readonly Patch[]` as there is no reason why it shouldn't accept it. If you think that other changes here are too risky (like returning `readonly Patch[]` from `produceWithPatches`) then I can revert those bits of this change.

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -941,7 +941,9 @@ function runBaseTest(name, autoFreeze, useStrictShallowCopy, useListener) {
 					if (canReferNonEnumerableProperty) s.foo.a++
 					if (useStrictShallowCopy) expect(isEnumerable(s, "foo")).toBeFalsy()
 				})
-				if (canReferNonEnumerableProperty) expect(nextState.foo).toBeTruthy()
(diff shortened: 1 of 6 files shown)


mweststrate:
Yeah that looks great! I'm not 100% sure about the returned patches indeed, as people might be postprocessing those in a mutable manner. So probably safer to not change right away. (Feel free to set up a separate PR to park for the next major)

mweststrate:
Looks like some of the tests fail, but overall agree with the idea :)

Andarist:
@mweststrate I adjusted the PR as requested. There are still some test failures here but they are unrelated to the change - the same ones can be seen `main`

mweststrate:
I didn't use GitHub for too long and can't find the rebase button anymore, but master should be fixed now.

Andarist:
Just rebased this on top of `main` - it should be ready to go

mweststrate:
Merging, thanks for the PR and sorry for the delay!

PR #1113, opened 2024-03-30
Fix typo in comment in common.ts

frwdrik:


src/utils/common.ts
```diff
@@ -62,7 +62,7 @@ export function original(value: Drafted<any>): any {
 
 /**
  * Each iterates a map, set or array.
- * Or, if any other kind of of object all it's own properties.
+ * Or, if any other kind of object, all of it's own properties.
```
nycdotnet, line 65:
```suggestion
 * Or, if any other kind of object, all of its own properties.
```

frwdrik: Even better!

PR #1124, opened 2024-05-18
fix: Mark exports as pure, for better tree-shakability

steveluscher:
Consider an application that uses _only_ `produce` but no other methods. This PR allows such an application to be properly tree-shaken by a bundler, to eliminate the other exports altogether.

```ts
import { produce } from "immer";
const state = {
  name: "A",
  icon: "III",
  accounts: [
    { address: "X", foo: "bax" },
    { address: "Y", foo: "bax" },
    { address: "Z", foo: "bax" },
  ],
};
const updateAccount = produce(
  (draft: typeof state, newAccount: (typeof state)["accounts"][number]) => {
    const accountIndex = draft.accounts.findIndex(
      ({ address }) => address === newAccount.address
    );
    if (accountIndex !== -1) {
      draft.accounts[accountIndex].address = newAccount.address;
      draft.accounts[accountIndex].foo = newAccount.foo;
    }
  }
);
const nextState = updateAccount(state, { address: "X", foo: "bar" });
console.log(state, nextState);
```

Bundle

```shell
pnpm dlx esbuild --bundle --define:process.env.NODE_ENV='"production"' --tree-shaking  index.ts
```

Diff as compared to the version on `main`

```diff
658,663d657
<   var produceWithPatches = immer.produceWithPatches.bind(immer);
<   var setAutoFreeze = immer.setAutoFreeze.bind(immer);
<   var setUseStrictShallowCopy = immer.setUseStrictShallowCopy.bind(immer);
<   var applyPatches = immer.applyPatches.bind(immer);
<   var createDraft = immer.createDraft.bind(immer);
<   var finishDraft = immer.finishDraft.bind(immer);
```

### DIFF
--- a/src/immer.ts
+++ b/src/immer.ts
@@ -45,13 +45,13 @@ const immer = new Immer()
  * @param {Function} patchListener - optional function that will be called with all the patches produced here
  * @returns {any} a new state, or the initial state if nothing was modified
  */
-export const produce: IProduce = immer.produce
+export const produce: IProduce = /* @__PURE__ */ immer.produce
 
 /**
  * Like `produce`, but `produceWithPatches` always returns a tuple
  * [nextState, patches, inversePatches] (instead of just the next state)
  */
-export const produceWithPatches: IProduceWithPatches = immer.produceWithPatches.bind(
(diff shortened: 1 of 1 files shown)


mweststrate (approved):
Thanks for catching this!

PR #1131, opened 2024-06-25
chore: added update-pattern on how to replace & bulk update (nested) objects

matt-miro:
Add an update pattern that explains how to replace & update objects in bulk through mutation, the docs felt like they could use a little more explanation

Kudos to @markerikson for offering guidance and input

### DIFF
--- a/website/docs/update-patterns.md
+++ b/website/docs/update-patterns.md
@@ -46,7 +46,7 @@ const updatedTodosObj = produce(todosObj, draft => {
 })
 ```
 
-Please keep in mind that when updating (nested) objects you should do so through mutation. Creating a new object and directly assigning it to a property in your immer draft object, will not work. This creates a new reference and does not update the existing object.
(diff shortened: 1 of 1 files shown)


website/docs/update-patterns.md
```diff
@@ -35,8 +35,19 @@ const deletedTodosObj = produce(todosObj, draft => {
...
+const updatedTodosObj = produce(todosObj, draft => {
+	Object.assign(draft, {
+		id1: {done: true, body: "Take out the trash"},
+		id2: {done: true, body: "Check Email"},
+		id3: {done: true, body: "Feed my cat"}
+	})
+})
 ```
 
+Please keep in mind that when updating (nested) objects you should do so through mutation. Creating a new object and directly assigning it to a property in your immer draft object, will not work. This creates a new reference and does not update the existing object.
```
markerikson, line 49: :hand: Not sure this phrasing is right.

Immer ultimately tracks mutations to the draft and returns a new reference if anything really did change.

`draft.someField = {}` _is_ a mutation, and it _will_ end up returning a new result reference.  This is good.

matt-miro: Discussed with Mark in private, replaced the note

Issue #1144, opened 2024-09-10
TypeScript error: 'key' is declared but its value is never read


alissalkvc:
Hi there, I'm using Immer in my TS project and I'm getting a TS error for the following line:
https://github.com/immerjs/immer/blob/e2d222bd4fb26abded04075c936290715e9ee335/src/utils/common.ts#L207

Running `tsc --noEmit` gives me this error:
```
node_modules/immer/src/utils/common.ts:207:33 - error TS6133: 'key' is declared but its value is never read.
```
is it possible for this to be fixed? Thanks.

markerikson:
Your project shouldn't be trying to compile Immer's own source in the first place. Can you check your TS config and figure out why it's trying to do that?

alissalkvc:
Sure. Here's the `tsconfig.json`. 

Immer was loaded into my project as a dependency of another third-party package. Anything jump out to you as part of this configuration?

```
{
  "compilerOptions": {
    "allowJs": true,
    "allowSyntheticDefaultImports": true,
    "alwaysStrict": true,
    "declaration": true,
    "emitDecoratorMetadata": true,
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "forceConsistentCasingInFileNames": true,
    "importHelpers": true,
    "incremental": true,
    "isolatedModules": true,
    "jsx": "react-jsx",
    "lib": ["dom", "dom.iterable", "esnext"],
    "module": "esnext",
    "noFallthroughCasesInSwitch": true,
    "noImplicitAny": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "moduleResolution": "node",
    "noEmit": true,
    "removeComments": true,
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "strictBindCallApply": false,
    "strictNullChecks": true,
    "target": "esnext",
    "typeRoots": ["./src/@types", "./node_modules/@types"]
  },
  "exclude": ["./node_modules", "../typings"],
  "include": ["./src", "./vite.config.ts", "./jest.config.ts"]
}
```

markerikson:
Hmm. Not immediately. `include`, `exclude`, and `skipLibCheck` all seem reasonable.

alissalkvc:
Perhaps because it's a transitive dependency in my project, and TS is type-checking it as part of the third-party module I'm using. I'll keep looking into it.

alissalkvc:
[...]

PR #1145, opened 2024-09-12
Prefix param with underscore to prevent unused param warning

alissalkvc:
Here's the current issue:
image

Fixes https://github.com/immerjs/immer/issues/1144 by prefacing the unused param with `_`.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -204,7 +204,7 @@ export function freeze<T>(obj: any, deep: boolean = false): T {
 	if (deep)
 		// See #590, don't recurse into non-enumerable / Symbol properties when freezing
 		// So use Object.entries (only string-like, enumerables) instead of each()
-		Object.entries(obj).forEach(([key, value]) => freeze(value, true))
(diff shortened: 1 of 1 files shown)


src/utils/common.ts
```diff
@@ -204,7 +204,7 @@ export function freeze<T>(obj: any, deep: boolean = false): T {
 	if (deep)
 		// See #590, don't recurse into non-enumerable / Symbol properties when freezing
 		// So use Object.entries (only string-like, enumerables) instead of each()
-		Object.entries(obj).forEach(([key, value]) => freeze(value, true))
+		Object.entries(obj).forEach(([_key, value]) => freeze(value, true))
```
mweststrate, line 207: Looks like this could be just `Object.values` which is probably faster and also skips the arg destructuring :)

mweststrate (changes requested):

src/utils/common.ts
alissalkvc: Updated!

mweststrate (approved):
This is great, thanks!

PR #1155, opened 2025-02-08
fix: unfinished todo filter condition

Ishrathh:


Devgametools (commented):
I found the same issue and this is the correct solution.

Ishrathh:
Glab to be contributing! Cheers 😄

PR #1156, opened 2025-03-04
docs(curry): first argument -> second argument

Haroenv:
I guess it's written first because the state is first, or otherwise a zero-based count?


mweststrate:
No, as shown in the example below, if you pass the function first, you cat a curried version of that function back

PR #1164, opened 2025-09-08
Optimize Immer performance where possible

markerikson:
## Overview

(Effectively stacked on #1162 for Vitest and #1163 for the perf benchmark harness, but keeping this separate for ease of review.)

Per #1152 , I've been investigating Immer's loss of performance over the last several major versions.

Using the perf benchmark harness and CPU profiling, I was able to use AI to help identify a series of hotspots and slow functions, then try out various approaches to optimize them and check the results.  The benchmark numbers are noisier than I'd like, but as best as I can tell all of these are an improvement individually, and together.

The additional good news is that the changes in this PR are fairly small tweaks (per-function caching or early bailouts), and not major architectural changes.  So, these should be feasible to review and merge, and also it means that a larger architectural change might have additional wins.

## Optimizations

This PR contains several optimizations:

- `isPlainObject`:  This turned out to be a surprising hotspot right away.  The culprit seems to be the repeated constructor stringification, which is almost always the same reference with plain objects.  Using a `WeakMap` to cache that eliminated `isPlainObject` as a hotspot.
- `finalizeProperty`: Adding some early returns avoided running the rest of the logic in cases like primitives.
- `isDraftable`: another `WeakMap` cache, plus adding a fast bailout path
- `each`: ported the logic from the `faster-iteration-experiment` branch that uses `Object.keys()` instead of `Reflect.ownKeys()`, and added a new `setUseStrictIteration` option.  Currently it defaults to `true` to maintain the existing behavior.

## Perf Results

If I run the benchmark with these changes, the relative results are:

```
  add: vanilla (freeze: true)
   1.01x faster than add: vanilla (freeze: false)
   23.59x faster than add: immer10Perf (freeze: true)
   32.58x faster than add: immer7 (freeze: false)
   35.79x faster than add: immer10 (freeze: false)
   36.23x faster than add: immer10Perf (freeze: false)
   39.12x faster than add: immer9 (freeze: false)
   42.62x faster than add: immer7 (freeze: true)
   43.01x faster than add: immer9 (freeze: true)
   44.97x faster than add: immer10 (freeze: true)
   173.72x faster than add: immer8 (freeze: false)
   185.97x faster than add: immer8 (freeze: true)
   
  remove: vanilla (freeze: true)
   1.03x faster than remove: vanilla (freeze: false)
   1279.15x faster than remove: immer7 (freeze: true)
   1318.47x faster than remove: immer10 (freeze: true)
   1450.17x faster than remove: immer9 (freeze: true)
[...]

PR #1183, opened 2025-10-15
Rewrite finalization system to use a callback approach instead of tree traversal

markerikson:
Stacked on top of #1164 for assorted smaller perf tweaks, followed by #1184 for array method overrides:

This PR ports Mutative's "finalization callback" approach as a more targeted and performant implementation for finalization compared to the existing recursive tree traversal approach:

- Added cleanup callbacks for each draft that's created
- Added callbacks to handle root drafts, assigned values, and recursing
  inside of plain values
- Updated state creation to return `[draft, state]` to avoid a lookup
- Rewrote patch generation system to work with callbacks instead of
  during tree traversal

It also makes some additional tweaks to perf and internal logic:

- Sets `strictIteration: false` (technically a breaking change) to get the benefits of faster loose iteration
- Updates `has / get / set` to allow passing `type? = getArchetype(value)` as the last argument. We know this in most places we call them, so we can pass it in without calling `getArchetype()` repeatedly
- Converts `assigned_` from a `Record` to a `Map` to enable size lookups
- Tweaks `freeze()` to remove an `isDraftable` check that wasn't necessary (and slightly expensive), and to use `each(val, cb, false)` for consistency with loose iteration

## Performance 

Stacked on top of the misc perf improvements from #1164 , and with `strictIteration: false`, this branch shows:

```
┌─────────────────────┬──────────────┬──────────────┬─────────────┐
│ Scenario            │ immer10      │ immer10Perf  │ Improvement │
├─────────────────────┼──────────────┼──────────────┼─────────────┤
│ remove-high-reuse   │        9.3ms │        2.8ms │      +70.0% │
│ mixed-sequence      │        3.3ms │        1.2ms │      +63.5% │
│ concat              │      132.4µs │       50.5µs │      +61.8% │
│ update-reuse        │        3.5ms │        1.7ms │      +52.0% │
│ update-high         │      105.5µs │       53.8µs │      +49.0% │
│ rtkq-sequence       │       14.8ms │        7.6ms │      +48.3% │
│ update-high-reuse   │        3.6ms │        2.0ms │      +44.6% │
│ updateLargeObject-r │       10.4ms │        6.2ms │      +40.2% │
│ update-multiple     │       49.2µs │       29.6µs │      +39.9% │
│ remove-reuse        │        4.6ms │        3.2ms │      +31.0% │
│ updateLargeObject   │      219.7µs │      158.3µs │      +28.0% │
│ update              │       24.8µs │       18.6µs │      +24.9% │
│ add                 │       26.2µs │       22.2µs │      +15.0% │
│ remove-high         │      118.3µs │      107.0µs │       +9.6% │
│ sortById-reverse    │      149.1µs │      144.0µs │       +3.4% │
│ remove              │      123.4µs │      144.3µs │      -16.9% │
│ reverse-array       │      110.6µs │      140.5µs │      -27.0% │
│ filter              │       50.1µs │       65.5µs │      -30.7% │
│ mapNested           │      129.6µs │      184.9µs │      -42.7% │
└─────────────────────┴──────────────┴──────────────┴─────────────┘

✓ immer10Perf shows an average 24.4% performance improvement over immer10
```

Since #1164 was ~20% faster than v10, this is another ~4.5% faster.

I'll note that's with some of the array scenarios actually showing regressions, so the other scenarios appear to be a bigger improvement than 4-5%.  I don't have a full explanation for why those scenarios got a bit slower.  Some of it may be doing additional work when creating proxies - I'd still like to dig into it a bit further.  However, the array scenarios are all in the microsecond range already, so not like it's a huge difference.  Additionally, the next architectural PR that adds overriding the array methods results in drastic improvements for all of those.

## Bundle Size

Eyeballing bundle sizes, this PR increases the `immer.production.js` minified bundle size by about 1700 bytes, from ~12K to ~14K.  If I build a Vite app and measure using Sonda, actual size in a built app appears to have grown from:

- v10.0.3 (which is out of date but what my branches were based from): 7.52K
- The original set of perf tweaks: 8.11K (which may include some of the other recent PRs that were merged - not sure my changes would have added that much)
- This PR with callbacks: 9.38K

I'm always very sensitive to increasing bundle size, so there's a very real question of whether adding a couple K here is worth it for a net 5-7% perf increase, especially when the larger benefits seem to be from the smaller tweaks and the array method overrides.

### DIFF
--- a/src/core/finalize.ts
+++ b/src/core/finalize.ts
@@ -108,9 +108,8 @@ function markStateFinalized(state: ImmerState) {
 	state.scope_.unfinalizedDrafts_--
 }
 
-function isSameScope(state: ImmerState, rootScope: ImmerScope) {
-	return state.scope_ === rootScope
-}
+let isSameScope = (state: ImmerState, rootScope: ImmerScope) =>
+	state.scope_ === rootScope
 
 // A reusable empty array to avoid allocations
 const EMPTY_LOCATIONS_RESULT: (string | symbol | number)[] = []
@@ -180,7 +179,7 @@ export function registerChildFinalizationCallback(
 		}
 
 		// Handle potential set value finalization first
-		rootScope.mapSetPlugin_?.fixPotentialSetContents(state)
+		rootScope.mapSetPlugin_?.fixSetContents(state)
 
 		const finalizedValue = getFinalValue(state)
 
--- a/src/core/immerClass.ts
+++ b/src/core/immerClass.ts
@@ -32,7 +32,9 @@ import {
 	AnySet,
 	isObjectish,
 	isFunction,
-	isBoolean
+	isBoolean,
+	PluginMapSet,
+	PluginPatches
 } from "../internal"
 
 interface ProducersFns {
@@ -122,7 +124,7 @@ export class Immer implements ProducersFns {
 			if (patchListener) {
 				const p: Patch[] = []
 				const ip: Patch[] = []
-				getPlugin("Patches").generateReplacementPatches_(base, result, {
+				getPlugin(PluginPatches).generateReplacementPatches_(base, result, {
 					patches_: p,
 					inversePatches_: ip
 				} as ImmerScope) // dummy scope
@@ -217,7 +219,7 @@ export class Immer implements ProducersFns {
 			patches = patches.slice(i + 1)
 		}
 
-		const applyPatchesImpl = getPlugin("Patches").applyPatches_
+		const applyPatchesImpl = getPlugin(PluginPatches).applyPatches_
 		if (isDraft(base)) {
 			// N.B: never hits if some patch a replacement, patches are never drafts
 			return applyPatchesImpl(base, patches)
@@ -239,9 +241,9 @@ export function createProxy<T extends Objectish>(
 	// returning a tuple here lets us skip a proxy access
 	// to DRAFT_STATE later
 	const [draft, state] = isMap(value)
-		? getPlugin("MapSet").proxyMap_(value, parent)
+		? getPlugin(PluginMapSet).proxyMap_(value, parent)
 		: isSet(value)
-		? getPlugin("MapSet").proxySet_(value, parent)
+		? getPlugin(PluginMapSet).proxySet_(value, parent)
 		: createProxyProxy(value, parent)
 
 	const scope = parent?.scope_ ?? getCurrentScope()
@@ -257,7 +259,7 @@ export function createProxy<T extends Objectish>(
 	} else {
 		// It's a root draft, register it with the scope
 		state.callbacks_.push(function rootDraftCleanup(rootScope) {
-			rootScope.mapSetPlugin_?.fixPotentialSetContents(state)
+			rootScope.mapSetPlugin_?.fixSetContents(state)
 
 			const {patchPlugin_} = rootScope
 
--- a/src/core/proxy.ts
+++ b/src/core/proxy.ts
@@ -18,7 +18,12 @@ import {
 	createProxy,
 	ArchType,
 	ImmerScope,
-	handleCrossReference
+	handleCrossReference,
+	WRITABLE,
+	CONFIGURABLE,
+	ENUMERABLE,
+	VALUE,
+	isArray
 } from "../internal"
 
 interface ProxyBaseState extends ImmerBaseState {
@@ -51,9 +56,9 @@ export function createProxyProxy<T extends Objectish>(
 	base: T,
 	parent?: ImmerState
 ): [Drafted<T, ProxyState>, ProxyState] {
-	const isArray = Array.isArray(base)
+	const baseIsArray = isArray(base)
 	const state: ProxyState = {
-		type_: isArray ? ArchType.Array : (ArchType.Object as any),
+		type_: baseIsArray ? ArchType.Array : (ArchType.Object as any),
 		// Track which produce call this is associated with.
 		scope_: parent ? parent.scope_ : getCurrentScope()!,
 		// True for both shallow and deep changes.
@@ -86,7 +91,7 @@ export function createProxyProxy<T extends Objectish>(
 	// Note that in the case of an array, we put the state in an array to have better Reflect defaults ootb
 	let target: T = state as any
 	let traps: ProxyHandler<object | Array<any>> = objectTraps
-	if (isArray) {
+	if (baseIsArray) {
 		target = [state] as any
 		traps = arrayTraps
 	}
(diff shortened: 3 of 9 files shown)


src/core/scope.ts
```diff
@@ -39,7 +41,9 @@ function createScope(
 		// Whenever the modified draft contains a draft from another scope, we
 		// need to prevent auto-freezing so the unowned draft can be finalized.
 		canAutoFreeze_: true,
-		unfinalizedDrafts_: 0
+		unfinalizedDrafts_: 0,
+		handledSet_: new WeakSet(),
```
mweststrate: since these scopes are thrown away at the end of the process, normal Sets should work and might maybe more efficient?

src/core/finalize.ts
```diff
@@ -43,163 +47,280 @@ export function processResult(result: any, scope: ImmerScope) {
...
-		let resultEach = result
-		let isSet = false
-		if (state.type_ === ArchType.Set) {
-			resultEach = new Set(result)
-			result.clear()
-			isSet = true
+		// Execute all registered draft finalization callbacks
+		if (state.callbacks_) {
+			while (state.callbacks_.length > 0) {
+				const callback = state.callbacks_.pop()!
```
mweststrate: I'm wondering if you can avoid the `pop` here. Not sure if `.length--` is faster? Or alternative, not modify the callbacks array at all, but separately keep a counter how for we got.

src/core/finalize.ts
```diff
@@ -43,163 +47,280 @@ export function processResult(result: any, scope: ImmerScope) {
...
+function markStateFinalized(state: ImmerState) {
+	state.finalized_ = true
+	state.scope_.unfinalizedDrafts_--
+}
+
+function isSameScope(state: ImmerState, rootScope: ImmerScope) {
+	return state.scope_ === rootScope
+}
+
+const NO_LOCATIONS: (string | symbol | number)[] = []
```
mweststrate: Comment on what this is?

src/core/proxy.ts
```diff
@@ -74,7 +72,9 @@ export function createProxyProxy<T extends Objectish>(
 		copy_: null,
 		// Called by the `produce` function.
 		revoke_: null as any,
-		isManual_: false
+		isManual_: false,
+		// `callbacks` actually gets assigned in `createProxy`
+		callbacks_: []
```
mweststrate: Q: would it help to pre-initialize a size? this captures all modifications for a single immer run? maybe 100 to avoid re-allocations on small recipes?

src/core/proxy.ts
```diff
@@ -116,7 +116,11 @@ export const objectTraps: ProxyHandler<ProxyState> = {
 		// Assigned values are never drafted. This catches any drafts we created, too.
 		if (value === peek(state.base_, prop)) {
 			prepareCopy(state)
-			return (state.copy_![prop as any] = createProxy(value, state))
+			// Ensure array keys are always numbers
+			const childKey = state.type_ === ArchType.Array ? Number(prop) : prop
+			const childDraft = createProxy(state.scope_, value, state, childKey)
+
+			return (state.copy_![prop as any] = childDraft)
```
mweststrate: `prop` => `childKey`?

src/core/immerClass.ts
```diff
@@ -220,17 +224,45 @@ export class Immer implements ProducersFns {
...
 		? getPlugin("MapSet").proxySet_(value, parent)
 		: createProxyProxy(value, parent)
 
 	const scope = parent ? parent.scope_ : getCurrentScope()
 	scope.drafts_.push(draft)
-	return draft
+
+	// Ensure the parent callbacks are passed down so we actually
+	// track all callbacks added throughout the tree
+	state.callbacks_ = parent?.callbacks_ ?? []
```
mweststrate: See earlier comment on initial callbacks size

src/core/finalize.ts
mweststrate: ah EMPTY_ARRAY_INSTANCE maybe?

src/core/finalize.ts
```diff
@@ -43,163 +47,280 @@ export function processResult(result: any, scope: ImmerScope) {
...
+function generatePatchesAndFinalize(
+	state: ImmerState,
+	patches?: Patch[],
+	inversePatches?: Patch[]
+) {
+	const shouldFinalize =
+		state.modified_ &&
+		!state.finalized_ &&
+		(state.type_ === ArchType.Set ||
+			(state.assigned_ && state.assigned_.size > 0))
```
mweststrate: if `assigned_` is always initialized (as currently the case), the left branch is always true? Otherwise `state.assigned_?.size > 0` as a NIT?

src/core/finalize.ts
```diff
@@ -43,163 +47,280 @@ export function processResult(result: any, scope: ImmerScope) {
...
+
+export function finalizeAssigned(
+	state: ImmerState,
+	key: PropertyKey,
+	rootScope: ImmerScope
+) {
+	const wasAssigned =
+		(state as Exclude<ImmerState, SetState>).assigned_!.get(key) ?? false
+
+	if (rootScope.drafts_.length > 1 && wasAssigned === true && state.copy_) {
```
mweststrate: NIT: wasAssigned looks like the most expensive condition, so maybe inline and put the copy_ check upfront?

src/core/finalize.ts
```diff
@@ -43,163 +47,280 @@ export function processResult(result: any, scope: ImmerScope) {
...
+		isDraft(target) ||
+		handledSet.has(target) ||
+		!isDraftable(target) ||
+		isFrozen(target)
+	) {
+		return target
+	}
+
+	if (!rootScope.immer_.autoFreeze_ && rootScope.unfinalizedDrafts_ < 1) {
+		// optimization: if an object is not a draft, and we don't have to
```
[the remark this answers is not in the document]
mweststrate: looks like a very cheap condition, maybe put entirely up front?

__tests__/base.js
```diff
@@ -2757,7 +2757,7 @@ function runBaseTest(name, autoFreeze, useStrictShallowCopy, useListener) {
 			})
 
 			// This actually seems to pass now!
-			it("cannot return an object that references itself", () => {
+			it.skip("cannot return an object that references itself", () => {
```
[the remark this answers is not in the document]
markerikson: The `.skip` was because this test no longer threw an error.

I did some double-checking and the behavior is now much more nuanced:

- Just doing `() => res.self` doesn't error.  
- self-references can show up in the final output...
- but as soon as you modify a value that is circular, that reference gets copied and isn't itself circular any more (ie its circular field now points to the original reference that _is_ circular)

I rewrote the test to reflect that behavior

src/core/finalize.ts
markerikson: Yep, renamed for clarity

mweststrate (approved):

mweststrate:
Looks great and I'm happy to merge this! The following would work, correct?

* Land this as Immer 10.0.0
* Land the array ops as 10.1.0 (no breaking changes, but significant improvement)
* Land the strict mode draft as 10.1.1 (no breaking changes? didn't study yet)

PR #1197, opened 2025-12-22
fix: recursive T for WritableDraft

unional:
This contains a proper fix for #990 

Added tests for the type check.

Included #1196 as I don't want to see the type errors in the tests.

### DIFF
--- a/__tests__/type-external.ts
+++ b/__tests__/type-external.ts
@@ -1,0 +1,19 @@
+import {isType, JSONArray, JSONObject, JSONTypes} from "type-plus"
+import {Draft} from "../src/types/types-external"
+
+describe("Draft<T>", () => {
+	test("can use JSONTypes as T", () => {
+		type A = Draft<JSONTypes>
+		isType.equal<true, JSONTypes, A>()
+	})
+
+	it("can use JSONArray as T", () => {
+		type A = Draft<JSONArray>
+		isType.equal<true, JSONArray, A>()
+	})
+
+	it("can use Tuple as T", () => {
+		type A = Draft<[string, number, JSONArray, JSONObject]>
+		isType.equal<true, [string, number, JSONArray, JSONObject], A>()
+	})
+})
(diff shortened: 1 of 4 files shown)


mweststrate:
Looks great, thanks!

wereHamster:
Looks like this change is causing the following code to fail:

```ts
function Test<A>() {
  produce(
    (draft) => {
      draft.prop = undefined; // <- type error here
    },
    { prop: undefined as undefined | A },
  );
}
```

Type error on indicated line:
> Type undefined is not assignable to type A | undefined extends infer V ? V extends object ? Draft<V> : V : never

unional:
> Looks like this change is causing the following code to fail:

I can reproduce it. It seems like the type cannot be resolved partially like this.

Is there a particular use case for this? Is there a reason to declare the type like this rather than as a whole like:

```ts
	function Test<A extends { prop: undefined }>() {
		produce(
			(draft) => {
				draft.prop = undefined;
			},
			{ prop: undefined } as A,
		);
	}
```

wereHamster:
My full code is more like this:

```tsx
function MultiLineChart<A extends TDatum<{}>>(props: Props<A>) {
  const [state, mutate] = useImmer({
    hovered: undefined as undefined | A,
  });

  const resetHover = React.useCallback(() => {
    mutate((draft) => {
      draft.hovered = undefined; // <- type error here
    });
  }, [mutate]);
}
```

Interestingly, the following (which is quite close to your code) also errors:

```ts
function Test<A>() {
  type S = { prop: undefined | A }

  produce(
    (draft) => {
      draft.prop = undefined; // <- type error here
    },
    { prop: undefined } as S,
  );
}
```

unional:
The issue is `S` is being distributed to `Draft<S>` and `S`.
Since TypeScript cannot determine the type of `A` during static analysis, it fails.

I have a fix to defer the analysis. Let me create a PR.

PR #1201, opened 2025-12-28
Fix broken array patching and ensure all values in draft Maps/Sets are finalized

### LINKED ISSUE #1199: Patching a nested array breaks in Immer 11 when a minifier renames fields
Per this RTK issue:

- https://github.com/reduxjs/redux-toolkit/issues/5159

A standard RTKQ cache update that updates a nested array fails.  We implement this via `produceWithPatches()`.  The user "mutates" the cache entry, we generate the patches, stick them in an action, and then apply the patches in a reducer:

```ts
let serverData = { lists: [{ id: 'list-1', items: [] }] };

const api = createApi({
  reducerPath: 'api',
  baseQuery: fakeBaseQuery(),
  endpoints: (builder) => ({
    getLists: builder.query({
      queryFn: async () => {
        return { data: serverData };
      },
    }),

    // ❌ Broken: mutates nested array
    addItemBroken: builder.mutation({
      queryFn: () => ({ data: null }),
      async onQueryStarted({ item }, { dispatch, getState }) {
        const patch = dispatch(
          api.util.updateQueryData('getLists', undefined, (draft) => {
            draft.lists[0].items.push(item); // broken
          })
        );

      logPatch(patch)
      },
    }),
```

The user reported that 2.9 generates this patch, which works correctly:

```json
{
  "op": "add",
  "path": [
    "lists",
    0,
    "items",
    0
  ],
  "value": {
    "id": "_nKnLLYGpTi1u6x7cG2ip",
    "label": "Item 0"
  }
}
```

But 11.0 generates this broken patch, which does not work:

```json
{
  "op": "add",
  "path": [
    0
  ],
  "value": {
    "id": "6Xna9j1k77dfw3ueIXFMD",
    "label": "Item 0"
  }
}
```

After investigation, I've concluded this is a problem with the logic in `getPath` in `patches.ts`:

```ts
	function getPath(state: ImmerState, path: PatchPath = []): PatchPath | null {
		// Step 1: Check if state has a stored key
		if ("key_" in state && state.key_ !== undefined) {
			// Step 2: Validate the key is still valid in parent

			const parentCopy = state.parent_!.copy_ ?? state.parent_!.base_
			const proxyDraft = getProxyDraft(get(parentCopy, state.key_!))
			const valueAtKey = get(parentCopy, state.key_!)
```

This works fine as long as the field `state.key_` actually exists.  However, if a minifier renames `.key_` to something else, the `if ("key_" in state` check will fail.

The users provided an Expo Snack that reproduced the issue ( https://snack.expo.dev/@jacobjaffe/a355a5 ).

If I download the snack and build it locally, the relevant minified `getPath` logic is:

```ts
// function createProxy()
      s.key_ = r,

// Patches plugin
          getPath: function e(t, n=[]) {
              if ("key_"in t && void 0 !== t.key_) {
                  const e = t.parent_.copy_ ?? t.parent_.base_
                    , r = C(M(e, t.key_))
                    , o = M(e, t.key_);
```

However, if I run the snack directly via the web UI, it still minifies the dependencies, and the relevant Immer `getPath` code becomes:

```ts
// function createProxy()
			(s.y = n),

// Patches plugin
			getPath: function t(r) {
				var n =
					arguments.length > 1 && void 0 !== arguments[1] ? arguments[1] : []
				if ("key_" in r && void 0 !== r.y) {
					var i,
						a = null != (i = r.i.e) ? i : r.i.t,
						u = B(F(a, r.y)),
						o = F(a, r.y)
```

Notice that `.key_` got renamed to `.y`!  That means the check will fail.

It's possible that just removing the `"key" in state` check will fix this.

### LINKED ISSUE #1200: Updated a nested Set in a Map leaks a `DraftSet` in Immer 11
Per this RTK issue:

- https://github.com/reduxjs/redux-toolkit/issues/5159

A user reported that they had a case where updating a nested `Set` wasn't working.  They linked a repro at https://github.com/keybase/client/pull/28706/files , which has:

```ts
import {enableMapSet} from 'immer'
import {create} from 'zustand'
import {immer} from 'zustand/middleware/immer'
enableMapSet()

type Reaction = {
  timestamp: number
  username: string
}
type ReactionDesc = {
  users: Set<Reaction>
}
type Reactions = ReadonlyMap<string, ReactionDesc>

type MinimalMessage = {
  conversationIDKey: string
  ordinal: number
  reactions?: Reactions
  type: 'text'
}

type Store = {
  messageMap: Map<number, MinimalMessage>
}

type State = Store & {
  toggleLocalReaction: (p: {
    decorated: string
    emoji: string
    targetOrdinal: number
    username: string
  }) => void
}

const initialOrdinal = 1
const initialMessage: MinimalMessage = {
  conversationIDKey: 'test-convo',
  ordinal: initialOrdinal,
  type: 'text',
}

const initialMessageMap = new Map<number, MinimalMessage>()
initialMessageMap.set(initialOrdinal, initialMessage)

export const useStore = create<State>()(
  immer((set, get) => ({
    messageMap: initialMessageMap,
    toggleLocalReaction: (p: {decorated: string; emoji: string; targetOrdinal: number; username: string}) => {
      const {emoji, targetOrdinal, username} = p
      set(s => {
        const m = s.messageMap.get(targetOrdinal)
        if (m) {
          const rs = {
            users: m.reactions?.get(emoji)?.users ?? new Set(),
          }
          if (!m.reactions) {
            m.reactions = new Map()
          }
          m.reactions.set(emoji, rs)
          const existing = [...rs.users].find(r => r.username === username)
          if (existing) {
            rs.users.delete(existing)
          }
          rs.users.add({timestamp: Date.now(), username})
        }
      })
    },
  }))
)
```

After turning that repro into a test, I do see a `DraftSet` leaking:

```
// logging
Message:  { id: 2, reactions: Map(1) { '👍' => { users: [DraftSet [Set]] } } }

 FAIL  __tests__/rtk-5159-zustand.ts > RTK #5159 - Zustand Immer Middleware (DEVELOPMENT mode) > Pattern 1: Map/Set Operations (Keybase pattern) > add reaction to message with existing reactions
Error: [Immer] Cannot use a proxy that has been revoked. Did you pass an object from inside an immer function to an async process? {}
 ❯ Module.die src/utils/errors.ts:45:9
     43|   const e = errors[error]
     44|   const msg = isFunction(e) ? e.apply(null, args as any) : e
     45|   throw new Error(`[Immer] ${msg}`)
       |         ^
     46|  }
     47|  throw new Error(
 ❯ assertUnrevoked src/plugins/mapset.ts:365:23
 ❯ DraftSet.values src/plugins/mapset.ts:278:4
 ❯ DraftSet.[Symbol.iterator] src/plugins/mapset.ts:306:16
 ❯ __tests__/rtk-5159-zustand.ts:139:24
```

markerikson:
This PR:

- Fixes an issue with `getPath` being broken in our pre-minified CJS/ESM production builds (but _not_ the main `immer.mjs` artifact, which is minified at app build time)
- Fixes an issue where a new plain object containing a draft value is inserted into a `Map` or `Set` and never gets finalized properly
- Rewrites `vitest.build.config.ts` to ensure that it does actually run against a real minified production artifact
- Rewrites the setup for `arrayTraps` to use `for..in` for setup instead of `each()`, to work around an odd issue with `each()` somehow not existing as `proxy.ts` is being executed in the production build
- Updates `package.json` to only rebuild the library when `test:build` is executed, as the previous use of `"pretest": "yarn build"` meant that it didn't get run if you were specifically executing `yarn test:build`. 

Fixes #1199
Fixes #1200 

## Background

After shipping RTK 2.11 with Immer 11, , a user reported an issue with RTKQ optimistic updates not working right when updating a nested array.  They also said this only happened in prod builds:

- https://github.com/reduxjs/redux-toolkit/issues/5159

After a bunch of discussion, we finally got a repro of that issue, and the repro showed that the patch contents had a very different path calculated when it failed.

Meanwhile, another user reported what they _thought_ was a related problem, but turned out to be completely different.  This involved updating nested `Map`s and `Sets`.

## Fixes

After investigation, I found that:

- Array patches not working: Immer's ESBuild config tries to minify "private" fields like `key_`, but the logic for patch path lookup had a `if ("key_" in state)` check.  That check failed when `key_` got renamed.
- Map/Set handling:  the repro takes an existing `DraftSet` instance, puts it in a new plain object, and puts the plain object into a `DraftMap`.  That effectively cuts the chain of assignments, and we never tried to fix up the `DraftSet` in the plain object.  We already have a `handleCrossReference` method that does this, it just wasn't being called inside of `DraftMap.set()` or `DraftSet.add()` (and also needed to be updated to handle working with `Set`s correctly)

Note that this meant that patches worked for _most_ apps, _if_ the build system picked up the normal `immer.mjs` artifact (which is unminified, has embedded `NODE_ENV` values, and will get minified by the app's build step).  The problem was if the build system picked our our pre-minified artifacts (`immer.cjs.production.js` or `immer.production.mjs`), where ESBuild had already mangled the field names.  Turns out that Expo does this, at least _some_ of the time.

In the process of working on this, I also found that my attempt to port the "production build" tests to Vitest wasn't working right - it wasn't actually loading the production artifact, so the tests were just running against the source code still.  I was able to fix that problem by updating the `alias` settings to be more specific.

While working on that, I also tweaked the `test:build` script to actually contain the build command.

Finally, while running the prod tests, I saw this very surprising error:

```
 FAIL  __tests__/map-set.js [ __tests__/map-set.js ]
TypeError: each is not a function
 ❯ src/core/proxy.ts:259:1
    257|
    258| const arrayTraps: ProxyHandler<[ProxyArrayState]> = {}
    259| each(objectTraps, (key, fn) => {
       | ^
    260|  // @ts-ignore
    261|  arrayTraps[key] = function() {
 ❯ src/internal.ts:25:31
```

I actually still have no explanation for why this is happening.  Looking at the prod artifacts, I see a minified version of `each()`, it exists in scope, and it ought to work fine.  But whatever, changing this to be a `for..in` loop works and passes tests.

### DIFF
--- a/__tests__/map-set.js
+++ b/__tests__/map-set.js
@@ -373,7 +373,7 @@ function runBaseTest(name, autoFreeze, useListener) {
 describe("RTK-5159: DraftSet leakage in Map", () => {
 	const {produce} = new Immer({autoFreeze: true})
 
-	test.only("DraftSet should not leak when placed in new object and set into Map", () => {
+	test("DraftSet should not leak when placed in new object and set into Map", () => {
 		const baseState = {
 			map: new Map([["key1", {users: new Set()}]])
 		}


mweststrate:
> Looking at the prod artifacts, I see a minified version of each(), it exists in scope, and it ought to work fine. 

Curious, was this minified to a `const` lambda? That might not hoist correctly like functions.

src/plugins/patches.ts
```diff
@@ -49,7 +49,7 @@ export function enablePatches() {
 
 	function getPath(state: ImmerState, path: PatchPath = []): PatchPath | null {
 		// Step 1: Check if state has a stored key
-		if ("key_" in state && state.key_ !== undefined) {
```
mweststrate: ugh, I should have spotted this in review

mweststrate:
Thanks for following up on these!

markerikson:
@mweststrate I see a `v11.1.1` tag here, but no release on NPM yet.  Need to kick the bot manually?

mweststrate:
Ok, scrambled the setup a bit more and triggered a new 11.1.2 release, that one should contain all changes now.

PR #1204, opened 2026-01-19
chore(deps): update prettier to 3.8.0 and pretty-quick to 4.2.2

unional:
The latest prettier vscode extension (v12) has drop support o prettier 1. It works with prettier 3.

I have added `"arrowParens": "avoid"` to `.prettierrc` to keep the style closer to the original.

There is also the `prettier-plugin-space-before-function-paren` plugin to keep the space before function paran. But I found that the code was not consistent on that to begin with so that actually produces more changes. So at the end I skip that.

The code change is added as a separate commit. So if further adjustments are needed, changes can be made and then drop that commit and apply a new one.


mweststrate:
Merged, thanks for updating!

PR #1215, opened 2026-03-24
chore: regression tests for fixed issue #916

oriondean:
Hey, first time contributor here! I've been a long term user (directly and indirectly) of immer so I'm really happy to contribute back.

I'm starting off with a relatively simple addition of regression tests verifying that issue #916 has been addressed.

This issue was originally fixed in [PR 917](https://github.com/immerjs/immer/pull/917), landing in `v9.0.19` of immer. You should find that bumping the version of immer to that version in the issue's [referenced codesandbox](https://codesandbox.io/s/minimal-repro-forked-xvb93p) will demonstrate it is no longer an issue.

### DIFF
--- a/__tests__/patch.js
+++ b/__tests__/patch.js
@@ -198,6 +198,69 @@ describe("applyPatches", () => {
 	})
 })
 
+describe("#916 nested patches with moved child draft", () => {
+	const makeBaseState = () => ({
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for adding the test, much appreciated!

Issue #831, opened 2021-07-17
Incorrect type inference with currying


michalenr:
## 🐛 Bug Report

When using currying, `produce` doesn't return the original type

## Link to repro

https://codesandbox.io/s/immer-sandbox-forked-342jd?file=/src/index.ts

## To Reproduce

Use `produce` function without data and check the return type.

## Observed behavior

Curried `produce` returns:
```ts
const updatedCache2: (oldAssignments: Assignment[]) => <Base extends readonly {
    readonly assignmentBk: string;
    readonly contentBk: string;
    readonly createdDate: Date;
    readonly startDate: Date;
    readonly dueDate: Date;
    readonly done: boolean;
}[]>(base?: Base) => Base
```

## Expected behavior

Curried `produce` returns the original type.

I need it, because it fails when using with Ramda's `pipe` function.

## Environment

We only accept bug reports against the latest Immer version.

- **Immer version:** 8.0.1
- [x] I filed this report against the _latest_ version of Immer
- [x] Occurs with `setUseProxies(true)`
- [ ] Occurs with `setUseProxies(false)` (ES5 only)

childrentime:
For the function `updateCache2` , arguments will be narrow to `Immutable<Assignment[]>`.  If we don't want this, we can specify the generic state. However, type of function `updatedCache3`  is `(state: Assignment[] | undefined) => Assignment[]` .Maybe this '?' can be removed？ 
```typescript
<State>(
  recipe: (
	  state: Draft<State>,
	  initialState: State
              ) => ValidRecipeReturnType<State>
): (state?: State) => State
```

childrentime:
it seems like this function 
```typescript
const updatedCache3 = produce<Assignment[]>((draft) => {
  const updatedAssignment = draft.find(
    ({contentBk}) => contentBk === assignment.contentBk
  )
  if (updatedAssignment) {
    updatedAssignment.done = !updatedAssignment.done
  }
})
```
will map to 
```typescript 
 <State>(recipe: (state: Draft<State>, initialState: State) => ValidRecipeReturnType<State>): (state?: State) => State;
```
not map to
```typescript
  <State>(recipe: (state: Draft<State>) => ValidRecipeReturnType<State>): (state: State) => State;
```

mweststrate:
Closing as fixed by #1216

Issue #877, opened 2021-12-03
[TS] IProduct (external-types.d.ts) curried overloading incorrect -> typecheck pass, runtime fail


akomm:
## 🐛 Bug Report

There is at least one cases of curried `produce` invocation with provided types, that pass type-check, but fail at runtime. The returned curried function incorrectly accepts `undefined` as `State` type-wise, but at runtime the recipe code will fail, when `undefined` is passed as state. This is because there are two overloading, of which one will always win due to the variance in the context.

I'v wrote explanation side-by-side with code in Sandbox.

## Link to repro

[Codesandbox](https://codesandbox.io/s/vigorous-mcclintock-kwn8j?file=/src/index.ts)
I'v wrote explanation side-by-side with code.

## To Reproduce

Open the Codesandbox and check open the codesandbox's console (tab underneath the preview)

## Observed behavior

Curried function returned from `produce`does accepts `undefined` as state.

## Expected behavior

Curried function returned from `produce`does not accept `undefined` as state.

## Environment

We only accept bug reports against the latest Immer version.

- **Immer version: 9.0.7**
- [X] I filed this report against the _latest_ version of Immer
- [X] Occurs with `setUseProxies(true)`
- [X] Occurs with `setUseProxies(false)` (ES5 only)

mweststrate:
Interested in attempting a PR?

akomm:
@mweststrate I would. I had not time to figure out what scenario the overloading `A` (as commented in the codesandbox) is referring to. Did not find any match in the docs.

mweststrate:
Closing as fixed by #1216

PR #1216, opened 2026-03-24
fix: remove invalid curried producer type

oriondean:
The removed curried producer type aims to resolve issues #831 and #877.

There's a couple of issues with the type, hence its removal in favour of the existing types:
 - It expects an `initialState` argument to be passed **within** the recipe itself, which isn't a valid use case
 - It allows state to be `undefined` even if the user has passed in a state generic type, which loosens inference


mweststrate:
This looks more accurate indeed, thanks!

PR #1228, opened 2026-04-07
fix: improve DraftMap.{entries,values}() compatibility

mrcljx:
I noticed two issues with `draftMap.entries()` (and `values()`):

- It doesn't support ES2025 features.
- It doesn't respect the contract that it should return itself for `[Symbol.iterator]()` but instead returns a new iterator instance (starting from the beginning again).

This PR addresses them both while **not** adding a hard dependency on ES2025 features.


mweststrate:
Looking great, thanks for making this contribution!

PR #1235, opened 2026-05-08
fix: reassigning to same value generates new copy with empty patches

### LINKED ISSUE #858: Reassigning to the same Date object generates new copy with empty patches
## 🐛 Bug Report

If you mutate `Date` property inside `produce`  but then revert it, immer still assumes changes while returning an empty set of patches.

## Link to repro

https://codesandbox.io/s/immer-reassigning-breaks-immutabillity-2u7ob

```ts

import {enablePatches, produceWithPatches} from "immer"

enablePatches();

type DataObject = { 
  date: Date;
}

let initialDate = new Date();
let data: DataObject[] = [{date: initialDate}];

let [newData, patches] = produceWithPatches(data, draft => {
  let element = draft[0];
  let dateBefore = element.date;
 
  element.date = new Date()
  element.date = dateBefore;
});

console.log("Patches:", patches) // []
console.log("New Date equal to Initial", newData[0].date === initialDate) // true
console.log("Array equal", newData === data); // expect true but getting false
```

## Environment

**Immer version: 9.0.6**

oriondean:
Resolves #858 

Temporary assignments that were later restored could still be treated as modifications, producing patches and creating new objects/arrays. This change reduces noise and work by detecting those revert-to-base scenarios early and short-circuiting them.

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -179,7 +179,7 @@ function runBaseTest(
 				delete s.a
 				s.a = a
 			})
-			expect(nextState).not.toBe(baseState)
+			expect(nextState).toBe(baseState)
 			expect(nextState).toEqual(baseState)
 		})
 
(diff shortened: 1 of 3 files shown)


mweststrate:
This is cool! But also a quite fundamental change in semantics and potential some perf ramifications. I'll have to think this one a bit further through :)

mweststrate:
TODO: check perf ramifications

PR #1236, opened 2026-05-08
fix: update current and original typings to assert draft value is unwrapped

### LINKED ISSUE #896: Typings of `current()` don't reflect that returned object is not a draft
## 🐛 Bug Report

The typings of `current()` state that the input and return types are the same. This doesn't reflect that the function unwraps drafts.

## Link to repro

```javascript
const draft = immer.createDraft({ a: true });
immer.current(draft);
```

## To Reproduce

Observe typings of the returned object. Runtime object is not a draft, but typings indicate that it is.

## Observed behavior

Return type is the same as the input type.

`export declare function current<T>(value: T): T;`

## Expected behavior

Return type is the type of thing whose draft was input.

`export declare function current<T>(value: Draft<T>): T;`

## Environment

- **Immer version:** 9.0.12

oriondean:
Resolves #896 

`current` and `original` typings have been updated to ensure that the passed in value is a `draft` type, with the returned type being the unwrapped (non-draft) equivalent.

### DIFF
--- a/__tests__/type-external.ts
+++ b/__tests__/type-external.ts
@@ -1,5 +1,6 @@
 import {isType, JSONArray, JSONObject, JSONTypes} from "type-plus"
 import {Draft} from "../src/types/types-external"
+import {createDraft, current, original} from "../src/immer"
 
 describe("Draft<T>", () => {
 	test("can use JSONTypes as T", () => {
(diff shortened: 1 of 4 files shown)


mweststrate:
Looking great, thanks!

PR #1239, opened 2026-05-11
docs: add typed custom produce function

### LINKED ISSUE #898: How to type a custom produce function?
## 🙋‍♂ Question

Hey folks. I'm using Immer with React and I want to create a custom produce function that does something with the new state. I want to pass this produce function down to components so they can call it to update the state, and so that I can do something with the newly updated state. I can't seem to figure out how to type that function however. It seems that `ValidRecipeReturnType` isn't exported. I was able to do it with `use-immer`'s `DraftFunction` but `use-immer` doesn't return the new state after it's set.

## Link to repro

https://codesandbox.io/s/festive-browser-v2wj8?file=/src/App.tsx

How can I type the `recipe` param from `updateState`?

## Environment

Latest React and Immer.

Thank you!

oriondean:
Resolves #898 

A small addition to docs demonstrating TypeScript and React typed custom prouce functions.

### DIFF
--- a/website/docs/produce.mdx
+++ b/website/docs/produce.mdx
@@ -81,6 +81,67 @@ expect(nextState[0]).toBe(baseState[0])
 expect(nextState[1]).not.toBe(baseState[1])
 ```
 
+## TypeScript example
+
+If you want to pass a typed recipe callback around, you can type it using `Draft` and allow either `void` or a replacement state as the return value.
+
+```ts
+import {produce} from "immer"
+import type {Draft} from "immer"
+
+type Recipe<S> = (draft: Draft<S>) => void | S
+
+function updateState<S>(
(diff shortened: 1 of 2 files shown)


mweststrate (changes requested):
First of all, thanks for adding this pattern! Secondly: sorry for the late response. 

Would you mind moving this pattern to typescript.md instead (https://immerjs.github.io/immer/typescript)?

Feel free to add a link to produce.md though to highlight that advanced typing scenarios are covered there, that might now not immediately obvious indeed.

oriondean:
No worries - I've moved the pattern across to `typescript.md` now and kept a cross-link. FYI - `pretty-quick` seems to have jumped in and tweaked some tags further up the doc that i've just left in.

mweststrate:
Merged, thanks!

Issue #1160, opened 2025-08-15
Unexpected undefined not assigned


Lin-coln:
## 🐛 Bug Report

When I assign "undefined" to the prop "foo" of draft, and the prototype of base has the prop "foo" which is "undefined",
the final return value is not assigned an "undefined" to the prop "foo".

## To Reproduce

```ts
import { enablePatches, immerable, produceWithPatches } from "immer";

const proto = { [immerable]: true, name: undefined };
const foo = Object.create(proto);

console.log(`[hasOwnProp] foo:`, Object.prototype.hasOwnProperty.call(foo, "name"));

enablePatches();
const [foo_next, patches, _] = produceWithPatches(foo, (x) => {
  console.log(`[immer] produce foo_next from immer`);
  x.name = undefined;
});

console.log(`[immer] foo_next patches:`, patches);

console.log(`[hasOwnProp] foo:`, Object.prototype.hasOwnProperty.call(foo, "name"));
console.log(`[hasOwnProp] foo_next:`, Object.prototype.hasOwnProperty.call(foo_next, "name"));

foo.name = undefined;
console.log(`[vanilla] assign name manually`);
console.log(`[hasOwnProp] foo:`, Object.prototype.hasOwnProperty.call(foo, "name"));
```

## Observed behavior

```bash
[hasOwnProp] foo: false
[immer] produce foo_next from immer
[immer] foo_next patches: []
[hasOwnProp] foo: false
[hasOwnProp] foo_next: false
[vanilla] assign name manually
[hasOwnProp] foo: true
```

## Expected behavior

```bash
[hasOwnProp] foo: false
[immer] produce foo_next from immer
[immer] foo_next patches: [
  {
    op: "add",
    path: [ "name" ],
    value: undefined,
  }
]
[hasOwnProp] foo: false
[hasOwnProp] foo_next: true
[vanilla] assign name manually
[hasOwnProp] foo: true
```

Lin-coln:
Adjusting the code here may solve it

`src/core/proxy.ts`

<img width="400" height="249" alt="Image" src="https://github.com/user-attachments/assets/27e40df0-4fd4-474a-bb96-48df84922919" />

replace `prop in state.copy_` to `has(state.copy_, prop)`.
In this way, the corresponding prop on the prototype will not be detected

PR #1262, opened 2026-07-02
fix: undefined assigned to a prototype-inherited key gets dropped

chatman-media:
Ran into this via #1160. If a draft's prototype has a key sitting at `undefined` (a class field with no initializer, or anything made with `Object.create`), assigning `undefined` to that same key on the draft is silently swallowed — no own property shows up on the result and no patch gets generated.

The `set` trap uses `prop in state.copy_` to tell "already has this key" apart from "brand new key," but `in` walks the prototype chain, so an inherited key looks like it's already there even though it isn't an own property yet. Swapped it for the `has()` own-property check that's already used a few lines above for basically the same reason.

Added a test in regressions.js that checks both `hasOwnProperty` on the result and the patch that comes out of `produceWithPatches`.

### DIFF
--- a/__tests__/regressions.js
+++ b/__tests__/regressions.js
@@ -5,10 +5,12 @@ import {
 	original,
 	isDraft,
 	immerable,
-	enableMapSet
+	enableMapSet,
+	enablePatches
 } from "../src/immer"
 
 enableMapSet()
+enablePatches()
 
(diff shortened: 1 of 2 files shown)


mweststrate:
Looking good, thanks for noticing and fixing!

PR #1267, opened 2026-07-09
fix: stabilize constructor draft wrapper

### LINKED ISSUE #1266: Regression when comparing equality using `fast-deep-equal`
## 🐛 Bug Report

`immer@11.1.9` regresses the identity of `draft.constructor` after the CVE fix for `constructor.prototype` / `__proto__` access. The security fix is understandable, but the implementation now returns a fresh wrapper on each access, which changes equality behavior for equal drafts inside a producer.

## To Reproduce

1. Install `immer@11.1.9`.
2. Run the following test:

```js
import assert from 'node:assert/strict';
import test from 'node:test';
import { produce } from 'immer';
import deepEqual from 'fast-deep-equal';

test('equal Immer drafts stay deep-equal', () => {
  produce({ a: { x: 1 }, b: { x: 1 } }, (draft) => {
    assert.equal(deepEqual(draft.a, draft.b), true);
    assert.equal(draft.a.constructor === draft.b.constructor, true);
  });
});
```

3. Compare the result with `immer@11.1.8`.

## Observed behavior

With `immer@11.1.9`, equal drafts no longer expose a stable `constructor` identity:

- `deepEqual(draft.a, draft.b)` can fail where it previously passed
- `draft.a.constructor === draft.b.constructor` is `false`

This is caused by the `constructor` proxy wrapper added in the CVE fix.

## Expected behavior

Equal drafts should preserve stable observable behavior for normal property reads, including `constructor`, while still blocking prototype pollution via `constructor.prototype` and `__proto__`.

## Environment

- **Introduced in immer version:** `11.1.9` (also happens with `11.1.11`)

## Notes

I validated that:

- `11.1.8` passes the repro
- `11.1.9` fails the repro
- a narrow fix that caches the constructor wrapper per underlying constructor restores the behavior while keeping the pollution guard in place

olemartinorg:
Fixes #1266
Code written by codex

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -2953,6 +2953,18 @@ function runBaseTest(
 			expect(result.objConstructed).toEqual(new Object().constructor(1))
 		})
 
+		it("keeps constructor wrappers stable for equal drafts", () => {
+			produce(
+				{
+					left: {x: 1},
+					right: {x: 1}
+				},
+				draft => {
+					expect(draft.left.constructor).toBe(draft.right.constructor)
+				}
+			)
+		})
+
 		it("does not allow prototype pollution via reserved constructor access", () => {
 			const pollutedKey = "__immer_test_polluted__"
 			const original = Object.prototype[pollutedKey]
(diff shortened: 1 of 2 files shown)


olemartinorg:
Please close this if #1269 is accepted as the solution :raised_hands:

mweststrate:
thanks nonetheless for putting up a pR!

Issue #1265, opened 2026-07-07
TypeError: 'get' on proxy: property 'prototype' is a read-only and non-configurable data property on the proxy target


dwiyatci:
We're upgrading from v11.1.7 to v11.1.11, and got the aforementioned error.

Here's the minimal repro example that resembles our code:
```js
import { produce } from "immer";
import _ from "lodash";

const baseState = [
  { title: "Learn Macroeconomics", done: false },
  { title: "Try Immer", done: false },
];

const nextState = produce(baseState, (draftState) => {
  draftState.push({ title: "Tweet about it" });
  draftState[1].done = true;

  if (
    _.isEqual(draftState, [
      { title: "Learn Macroeconomics", done: false },
      { title: "Try Immer", done: true },
      { title: "Tweet about it" },
    ])
  ) {
    console.log("yay!");
  }
});

console.log({ baseState, nextState });
```

It throws:
```
TypeError: 'get' on proxy: property 'prototype' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected '#<Object>' but got '[object Object]')
    at isPrototype (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:6475:54)
    at baseKeys (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:3530:12)
    at keys (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:13406:60)
    at baseGetAllKeys (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:3099:20)
    at getAllKeys (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:5947:14)
    at equalObjects (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:5865:22)
    at baseIsEqualDeep (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:3380:14)
    at baseIsEqual (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:3321:14)
    at equalArrays (/Users/Glenn/node_modules/.pnpm/lodash@4.18.1/node_modules/lodash/lodash.js:5757:17)
[...]

PR #1269, opened 2026-07-09
fix: return real constructor on drafts (fixes lodash.isEmpty crash), keep __proto__ pollution guard

### LINKED ISSUE #1268: lodash.isEmpty() on a draft throws "TypeError: 'get' on proxy: property 'prototype'..." since 11.1.9
## 🐛 Bug Report

Since **11.1.9** (still present in **11.1.11**), reading `constructor` on a draft returns a wrapper `Proxy` whose `prototype` read returns a fabricated value. Because a function's `prototype` is a non-configurable, non-writable data property, returning anything else violates a Proxy invariant and throws:

```
TypeError: 'get' on proxy: property 'prototype' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected '#<Object>' but got '[object Object]')
```

Any ecosystem code that reads `value.constructor.prototype` on a draft now crashes. The most common trigger is **`lodash.isEmpty()` on a draft inside a Redux Toolkit reducer** (lodash reads `value.constructor.prototype`); it also hits `lodash.isPlainObject`, `constructor === Object` brand checks, and drafts. carrying a data key named `constructor`.

Introduced by [`48fc378`](https://github.com/immerjs/immer/commit/48fc378) ("prevent prototype pollution via constructor.prototype access", #1259).

## Link to repro

Minimal repository: https://github.com/jonatankruszewski/immer-lodash-isempty-repro (`npm install && npm run repro`).
You can clone and follow the README.md or do a small reproduction locally as follows:

## To Reproduce

1. `npm install immer@11.1.11 @reduxjs/toolkit@2.12.0 lodash@4.18.1`
2. Run:

```js
const {configureStore, createSlice} = require("@reduxjs/toolkit")
const _ = require("lodash")

const slice = createSlice({
	name: "demo",
	initialState: {settings: {options: {}}},
	reducers: {
		setOptions(state, action) {
			_.isEmpty(state.settings.options) // throws on immer >= 11.1.9
			state.settings.options = action.payload
		}
	}
})

const store = configureStore({reducer: {demo: slice.reducer}})
store.dispatch(slice.actions.setOptions({items: [{id: 1}]})) // TypeError
```

Dependency-free equivalent:

```js
const {produce} = require("immer")
produce({}, d => {
	d.constructor.prototype
}) // TypeError (Proxy invariant)
```

## Observed behavior

`lodash.isEmpty(draft)` (and any read of `draft.constructor.prototype`) throws the `TypeError` above instead of returning a value. `11.1.8` works; `11.1.9`+ throws.

This is faulty because the `constructor` guard added in 11.1.9 cannot actually protect anything — it only crashes. On a draft, `draft.constructor` **is** the real `Object` and `draft.constructor.prototype` **is** the real `Object.prototype`, so `draft.constructor.prototype.x = 1` is literally `Object.prototype.x = 1`, plain JavaScript reachable in any reducer with or without a draft. Immer is not the vector, and it cannot intercept this without
returning a fake `constructor` (which would break `constructor === Object`).

Measured across every pollution vector (`via immer` vs the identical operation in plain JS with no immer — `npm run probe` in the repo):

| Vector | 11.1.8 | 11.1.11 (guard) | plain JS |
| --- | --- | --- | --- |
| `draft.constructor.prototype.x = 1` | pollutes | **throws** | pollutes |
| `draft.__proto__.x = 1` | pollutes | clean | pollutes |
| `draft.__proto__ = {…}` | throws¹ | clean | pollutes |
| `applyPatches(__proto__ path)` | throws | throws | n/a |

¹ pre-existing `setPrototypeOf` trap.

The guard's only effect on the `constructor.prototype` vector is the crash. It also causes collateral regressions vs 11.1.8 (`npm run probe` in the repo shows this dependency-free):

| Behavior | 11.1.8 | 11.1.9+ |
| --- | --- | --- |
| `lodash.isEmpty(draft)` | works | **throws `TypeError`** |
| `draft.constructor.prototype` (read) | works | **throws `TypeError`** |
| `produce({constructor: 'x'}, d => d.constructor)` | works | **throws `TypeError`** |
| `draft.constructor === Object` | `true` | `false` |
| `'constructor' in draft` | `true` | `false` |
| `draft.prototype = 42` (own data key) | persisted | **silently dropped** |

## Expected behavior

Reading `constructor` / `constructor.prototype` on a draft should return the real values and not throw, so ecosystem inspection (`lodash.isEmpty`, `constructor === Object`) keeps working, while the genuine prototype-pollution protections are preserved.

I have a working fix (all 3733 tests pass): keep the `__proto__` sanitizing wrapper (which cleanly blocks `draft.__proto__.x = 1` and is invariant-safe because `__proto__` resolves to a plain object), and drop only the `constructor` wrapper and the `has`/`set` reserved-key blocks. `setPrototypeOf` and patch-layer
guards are untouched, so the result is strictly safer than 11.1.8. It adds tests that distinguish a clean block from a crash, including unit tests for the `__proto__` guard wrapper.

- Diff: https://github.com/immerjs/immer/compare/main...jonatankruszewski:immer:jony-demo

## Environment

- **Immer version:** `11.1.11` (first bad: `11.1.9`; `11.1.8` works)
- [x] Occurs with `setUseProxies(true)` (proxies are the only mode in v11)
- [ ] Occurs with `setUseProxies(false)` (ES5 only) — N/A, ES5 mode was removed in immer v10

jonatankruszewski:
Fixes #1268

## Problem

The prototype-pollution hardening in #1259 (`48fc378`) wraps `constructor` reads on a draft in a `Proxy`. Because `draft.constructor` is the real `Object`, whose `prototype` is a non-configurable, non-writable data property, reading `draft.constructor.prototype` violates a Proxy invariant and throws:

```
TypeError: 'get' on proxy: property 'prototype' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value
```

This crashes any ecosystem code that inspects `value.constructor.prototype` on a draft — most notably `lodash.isEmpty()` inside a Redux Toolkit reducer.

## Why the `constructor` wrapper can't work

On a draft, `draft.constructor` **is** the real `Object` and `draft.constructor.prototype` **is** the real `Object.prototype`, so `draft.constructor.prototype.x = 1` is literally `Object.prototype.x = 1` — plain JavaScript reachable in any reducer with or without a draft. Immer is not the vector, and cannot intercept it without returning a fake `constructor` (which breaks `constructor === Object`). The wrapper's only effect on that vector was to crash. No version ever actually *blocked* it — 11.1.8 pollutes, 11.1.9+ throws, plain JS pollutes.

## Change

- **Keep** the `__proto__` protection — but as a small dedicated wrapper (`protoGuardTraps`). `__proto__` resolves to a plain object, so it can be virtualized without hitting the invariant, and it still cleanly blocks `draft.__proto__.x = 1`.
- **Drop** the `constructor` wrapper — return the real constructor. Fixes the crash, restores `constructor === Object`, and stops `new Proxy(...)` from throwing on drafts that carry a non-object `constructor` data key.
- **Drop** the `has`/`set` reserved-key blocks — restore `'constructor' in draft` and stop silently dropping writes to own data keys named `constructor`/`prototype`.

`setPrototypeOf` and the patch-layer guards are untouched, so this is **strictly safer than 11.1.8** (it adds the `__proto__` block) with the crash removed.

## Tests

The three tests added in #1259 wrapped each attack in `try/catch` and asserted only that `Object.prototype` stayed clean, so a crash and a real block passed identically (which is how the regression reached a release). Replaced them with tests that distinguish the two:

- `constructor` / `constructor.prototype` reads don't throw and `=== Object`;
- own data keys named like reserved properties round-trip;
- `Object.setPrototypeOf(draft, …)` throws;
- `draft.__proto__.x = 1` does **not** pollute (asserts *blocked*, not *threw*);
- unit tests for the `__proto__` guard wrapper.

All existing tests pass.

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -2953,96 +2953,102 @@ function runBaseTest(
 			expect(result.objConstructed).toEqual(new Object().constructor(1))
 		})
 
-		it("does not allow prototype pollution via reserved constructor access", () => {
-			const pollutedKey = "__immer_test_polluted__"
-			const original = Object.prototype[pollutedKey]
+		it("does not break constructor/prototype inspection on drafts", () => {
+			produce({data: {}}, draft => {
+				expect(draft.constructor).toBe(Object)
+				expect("constructor" in draft).toBe(true)
+				expect(() => draft.constructor.prototype).not.toThrow()
+				expect(draft.constructor.prototype).toBe(Object.prototype)
+				const ctor = draft.data.constructor
+				const proto =
+					(typeof ctor === "function" && ctor.prototype) || Object.prototype
+				expect(draft.data === proto).toBe(false)
+			})
+		})
 
-			delete Object.prototype[pollutedKey]
+		it("preserves own data keys named like reserved properties", () => {
+			const next = produce({constructor: "hi", prototype: 1}, draft => {
+				draft.prototype = 42
+				draft.extra = "x"
+			})
+			expect(next).toEqual({constructor: "hi", prototype: 42, extra: "x"})
+		})
 
+		it("blocks setPrototypeOf / __proto__ assignment on a draft", () => {
+			expect(() =>
+				produce({}, draft => {
+					Object.setPrototypeOf(draft, {polluted: true})
+				})
+			).toThrowError(/setPrototypeOf/)
+		})
+
+		it("blocks prototype pollution via draft.__proto__ traversal", () => {
+			const key = "__immer_proto_pollution__"
+			delete Object.prototype[key]
 			try {
-				// The attack should either throw an error or silently fail
-				// but NOT pollute Object.prototype
-				let hadError = false
-				try {
-					produce({}, draft => {
-						draft.constructor.prototype[pollutedKey] = true
-						draft["__proto__"][pollutedKey] = true
-					})
-				} catch (e) {
-					// Expected: error when trying to set on undefined/frozen objects
-					hadError = true
-				}
-
-				// The critical check: Object.prototype must not be polluted
-				// either because we threw an error OR because the assignment was silently blocked
-				expect(Object.prototype[pollutedKey]).toBeUndefined()
+				produce({}, draft => {
+					draft.__proto__[key] = true
+					draft["__proto__"][key] = true
+				})
+				expect(Object.prototype[key]).toBeUndefined()
+				expect({}[key]).toBeUndefined()
 			} finally {
-				if (original === undefined) {
-					delete Object.prototype[pollutedKey]
-				} else {
-					Object.prototype[pollutedKey] = original
-				}
+				delete Object.prototype[key]
 			}
 		})
 
-		it("blocks prototype pollution via stored constructor reference (CVE bypass)", () => {
-			const pollutedKey = "__immer_test_ref__"
-			const original = Object.prototype[pollutedKey]
+		describe("__proto__ guard wrapper (protoGuardTraps)", () => {
+			it("forwards ordinary reads to the real prototype", () => {
+				produce({}, draft => {
+					const guard = draft.__proto__
+					expect(typeof guard.hasOwnProperty).toBe("function")
+					expect(typeof guard.toString).toBe("function")
+					expect(guard.hasOwnProperty).toBe(Object.prototype.hasOwnProperty)
+				})
+			})
 
-			delete Object.prototype[pollutedKey]
+			it("returns a frozen null-prototype object for prototype-chain reads", () => {
+				produce({}, draft => {
+					const guard = draft.__proto__
+					for (const key of ["__proto__", "prototype"]) {
+						const blocked = guard[key]
+						expect(Object.isFrozen(blocked)).toBe(true)
+						expect(Object.getPrototypeOf(blocked)).toBe(null)
+						expect(Object.keys(blocked)).toHaveLength(0)
+					}
+				})
+			})
 
-			try {
-				// Attack 3 from CVE: store constructor reference and mutate
-				let hadError = false
-				try {
-					produce({data: {}}, draft => {
-						const ctor = draft.data.constructor
-						ctor.prototype[pollutedKey] = true
-					})
-				} catch (e) {
-					hadError = true
-				}
-
-				expect(Object.prototype[pollutedKey]).toBeUndefined()
-			} finally {
-				if (original === undefined) {
-					delete Object.prototype[pollutedKey]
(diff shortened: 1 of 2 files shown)


jonatankruszewski:
This also resolves #1266 (unstable `draft.constructor` identity breaking `fast-deep-equal`), and does it more completely than #1267.

#1267 caches the wrapper so `draft.a.constructor === draft.b.constructor` is stable again — but a cached wrapper is still a `Proxy` over `Object`, so reading `wrapper.prototype` still hits the invariant and throws. The lodash crash from #1268 would remain:

```js
wrap(Object) === wrap(Object)  // true  (identity fixed by caching)
wrap(Object).prototype         // TypeError: 'get' on proxy: property 'prototype' ...
```

Returning the real `constructor` (this PR) fixes both at once: the real `Object` has stable identity for free (`Object === Object`), and `.prototype` reads work because there's no wrapping Proxy to violate the invariant. So this supersedes #1267 and closes #1266 as well.

jonatankruszewski:
This also fixes #1265. It's the same regression reached through a different lodash path: there `_.isEqual(draft, …)` goes `baseKeys` → `isPrototype`, which reads `value.constructor.prototype` — the exact access the `constructor` wrapper crashes on. Same error, same commit, same fix.

Verified against that issue's repro on this branch:

```js
const {produce} = require("immer")
const _ = require("lodash") // 4.18.1

produce(
  [
    {title: "Learn Macroeconomics", done: false},
    {title: "Try Immer", done: false}
  ],
  draft => {
    draft.push({title: "Tweet about it"})
    draft[1].done = true
    _.isEqual(draft, [
      {title: "Learn Macroeconomics", done: false},
      {title: "Try Immer", done: true},
      {title: "Tweet about it"}
    ]) // no longer throws
  }
)
```

mweststrate:
I think actually the prototype hardening really didn't make sense, so I just reverted it. Nonetheless thanks for investigating this! We might still need this if a convincing case is made we actually need the hardening.

jonatankruszewski:
@mweststrate Thanks to you for the new patch!

Would you be open for a PR adding test cases for prototype inspection?
That would help preserve current behavior on all 11.X immer versions and safe guard consumers who depend on it

mweststrate:
Sure!

Op ma 20 jul 2026, 19:26 schreef Jonatan Kruszewski <
***@***.***>:

mweststrate:
Closing as the original cause has been reverted.

PR #1272, opened 2026-07-20
test: pin draft prototype-inspection behavior restored by #1271

jonatankruszewski:
Follow-up to our conversation. This adds a test-only file that pins the prototype-inspection behavior of drafts, so it can't silently regress in a future 11.x refactor.

### Why

#1259 wrapped `constructor`/`__proto__` reads in a sanitizing Proxy. That broke ecosystem libraries which introspect prototypes to classify values — e.g. lodash's `isPrototype` (behind `_.isEmpty`/`_.isEqual`/`_.isPlainObject`) reads
`value.constructor.prototype`, and fast-deep-equal compares `a.constructor`.

Reading `constructor.prototype` through the guard tripped the Proxy invariant for `Object`'s non-configurable `prototype`, so those reads threw or misreported

(#1265, #1266, #1268). It was reverted in #1271.

Right now nothing in the suite locks in the reverted behavior, so the same guard could get reintroduced without a failing test. These cases close that gap.

### What's covered

- `draft.constructor` returns the real constructor for object, array, and `[immerable]` class drafts
- constructor identity stays stable across sibling drafts (#1266)
- `draft.constructor.prototype` is readable and doesn't throw (#1265, #1268)
- lodash's `isPrototype` idiom works on drafts (#1268)
- reserved names (`constructor`) report correctly through `in` / `Object.keys`, and own data keys named `constructor` / `prototype` round-trip
- `Object.getPrototypeOf` / `__proto__` return the real prototype
- `setPrototypeOf` on a draft is still forbidden

### Notes

- No source changes — `__tests__/prototype-inspection.js` only.
- The file has a header comment documenting the rationale (the #1259 → #1271 history and the two consumer patterns), so the intent survives future  refactors. Happy to trim it if you'd rather keep it lean.

### DIFF
--- a/__tests__/prototype-inspection.js
+++ b/__tests__/prototype-inspection.js
@@ -1,0 +1,93 @@
+"use strict"
+import {produce, immerable} from "../src/immer"
+
+// Ecosystem libraries introspect prototypes to classify values, and reducers
+// routinely run them against drafts. The patterns below are lifted from the
+// most-installed consumers so drafts stay safe for them:
+// - lodash `isPrototype` (used by _.isEmpty, _.isEqual, _.isPlainObject):
+//   reads `value.constructor.prototype`
+// - fast-deep-equal: compares `a.constructor !== b.constructor`
+// These once crashed or misreported on drafts (#1265, #1266, #1268) because a
+// guard wrapped `constructor` reads in a Proxy, violating the engine invariant
+// for Object's non-configurable `prototype` property. The guard was reverted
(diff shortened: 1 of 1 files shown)


jonatankruszewski:
@mweststrate can we merge this one?

mweststrate:
@jonatankruszewski yes! Thanks for putting this one up. 

For future reference, this is basically why we're not going to mess with prototypes: https://adventures.nodeland.dev/archive/no-we-cant-harden-nodejs-against-prototype/

PR #1274, opened 2026-07-22
docs: fix RFC-6902 JSON Pointer path conversion

### LINKED ISSUE #1051: Documentation gives incorrect advice for making patches RFC-6902 compliant
## 🐛 Bug Report

[The patches documentation](https://immerjs.github.io/immer/patches) says

> The generated patches are similar (but not the same) to the [RFC-6902 JSON patch  standard](https://datatracker.ietf.org/doc/html/rfc6902/#section-4.1), except that the path property is an array, rather than a string. This makes processing patches easier. If you want to normalize to the official specification, patch.path = patch.path.join("/") should do the trick.

The advice for making the patches RFC-6920 compliant is incorrect. The suggested adjustment, `patch.path.join("/")`, will never return a `path` string with a leading `/`.

[RFC-6902]( https://datatracker.ietf.org/doc/html/rfc6902/) requires the path to be a JSON Pointer, defined in [RFC-6901](https://datatracker.ietf.org/doc/html/rfc6901/). The grammar in RFC-6901 §3 defines a `json-pointer` token as follows:
```
 json-pointer    = *( "/" reference-token )
```
Thus, the only `json-pointer` which does not start with `/` is the empty string.

The suggestion in the docs produces non-compliant paths such as `a/b/c` (instead of `/a/b/c`).

## Link to repro

See next section.

## To Reproduce

```
const path = ['a'];
console.log(path.join("/"));
```

## Observed behavior

```
a
```

## Expected behavior

```
/a
```

Obviously we don't expect `join()` to produce this, but rather we expect the docs to suggest code which would.

## Environment

Any environment you want---this is how `join()` works.

- **Immer version:**

Hashim1999164:
## Summary
- Corrects the patches docs tip for converting Immer path arrays into RFC-6902 JSON Pointers.
- Uses a leading `/` (`"/" + patch.path.join("/")`) so the result matches RFC-6901 JSON Pointers.
- Notes token escaping for `~` and `/`, and updates the zh-CN translation to match.

Fixes #1051

## Test plan
- [ ] Confirm the English and zh-CN patches pages render the updated sentence
- [ ] Confirm `"/" + ["a","b"].join("/")` yields `/a/b` as described in the issue

### DIFF
--- a/website/docs/patches.mdx
+++ b/website/docs/patches.mdx
@@ -118,7 +118,7 @@ expect(state).toEqual({
 })
 ```
 
-The generated patches are similar (but not the same) to the [RFC-6902 JSON patch standard](https://datatracker.ietf.org/doc/html/rfc6902/#section-4.1), except that the `path` property is an array, rather than a string. This makes processing patches easier. If you want to normalize to the official specification, `patch.path = patch.path.join("/")` should do the trick. Anyway, this is what a bunch of patches and their inverse could look like:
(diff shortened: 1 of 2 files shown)


mweststrate:
thanks!

PR #1289, opened 2026-08-15
fix: preserve structural sharing for no-op array-methods calls

maximilliangrand:
### Problem

With `enableArrayMethods()` enabled, mutating array methods that change nothing still produce a new state reference, breaking Immer's structural-sharing guarantee that a change-free producer returns the base.

```js
enableArrayMethods()
const base = {list: [1, 2, 3]}
const next = produce(base, d => d.list.push(...[])) // empty spread
next === base // false  (expected: true)
```

Same for `splice(i, 0)`, and `pop()`/`shift()` on an empty array. The common real-world trigger is `draft.list.push(...incoming)` when `incoming` happens to be empty, which silently invalidates memoization and forces re-renders. Native Immer (plugin disabled) returns the base in all these cases.

### Cause

The plugin routes every mutating method through `executeArrayMethod`, which calls `markChanged` unconditionally, bypassing the no-op detection that core's `set` trap performs.

### Fix

Detect these no-ops in O(1) and skip marking the array changed, so the producer returns the base and emits no patches, matching core behavior. Real mutations and their patches are unchanged.

`sort()`/`reverse()` are left untouched: they are genuine reorder operations where cheap no-op detection isn't possible.

### Verification

- New tests run in both plugin-enabled and native modes; they fail on baseline (source reverted) and pass with the fix.
- Full src suite and build-config suite pass; `prettier --check` clean.

### DIFF
--- a/__tests__/base.js
+++ b/__tests__/base.js
@@ -342,6 +342,17 @@ function runBaseTest(
 						expect([...next].find(Array.isArray)).toBe(arr)
 					})
 
+					it("sort()/reverse() on arrays with fewer than two elements return the base", () => {
+						const base = {empty: [], single: [1]}
+						const next = produce(base, d => {
+							d.empty.sort()
+							d.empty.reverse()
+							d.single.sort((a, b) => a - b)
+							d.single.reverse()
+						})
+						expect(next).toBe(base)
+					})
+
 					it("no-op splice() on an array inside a Set preserves the array's identity", () => {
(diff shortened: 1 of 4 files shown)


markerikson:
The idea is right, but looks like this was still doing some mutations of the copy due to order of operations.  Added some tests to verify and fixed that.  Also threw in a fix for an adjacent bug I found with when to use the copied value.

PR #1290, opened 2026-08-17
fix: remove global var Iterator declaration conflicting with ESNext lib (#1273)

### LINKED ISSUE #1273: Conflict between TS global Iterator definition and ESNext lib
## 🐛 Bug Report

There is a conflict between the global Iterator type introduced in immer v11.1.11 and the default Iterator type in TypeScript's ESNext lib.

The whole error:
```
node_modules/immer/dist/immer.d.ts:214:9 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'Iterator' must be of type 'IteratorConstructor', but here has type '{ from<T, TReturn>(iterable: Iterator<T, TReturn, any>): IterableIterator<T>; } | undefined'.

214     var Iterator: undefined | {
            ~~~~~~~~

  node_modules/@typescript/typescript-linux-x64/lib/lib.es2025.iterator.d.ts:145:9 - 'Iterator' was also declared here.
    145     var Iterator: IteratorConstructor;
                ~~~~~~~~


Found 1 error in node_modules/immer/dist/immer.d.ts:214
```

## Link to repro

[https://github.com/OmniSliver/immerGlobalIterator](https://github.com/OmniSliver/immerGlobalIterator)

(I don't know how to run the TypeScript checker in codesandbox)

## To Reproduce

- npm ci
- npm run ts

## Observed behavior

I get an error when running the TypeScript checker using the ESNext lib option

## Expected behavior

To not get the error.

## Environment

- **Immer version:** v11.1.15, but introduced in v11.1.11
- [ ] Occurs with `setUseProxies(true)`

ErfanBagheri404:
## Summary

The `declare global { var Iterator }` shim added to feature-detect ES2025's `Iterator.from` leaks a global variable declaration into the bundled `dist/immer.d.ts`. Consumers compiling against the `esnext` lib hit TS2403 because the global `Iterator` now collides with the standard `IteratorConstructor`.

## Changes (packages/immer/src/plugins/mapset.ts)

- Removed the `declare global { var Iterator }` block.
- Added a module-scoped feature detect that reads the iterator constructor off `globalThis` (`_globalIterator`), so the emitted `.d.ts` no longer declares a conflicting global.
- `iteratorFrom` uses `_globalIterator!.from` when present, otherwise the existing manual fallback.

Runtime behavior is unchanged.

## Test

- `bun run build` succeeds; `dist/immer.d.ts` no longer contains a global `var Iterator`.
- Repro with `--lib esnext --noEmit` against `dist/immer`: no TS2403.
- `bun run test:src`: 3755 passed, 8 skipped.

Fixes #1273

### DIFF
--- a/src/plugins/mapset.ts
+++ b/src/plugins/mapset.ts
@@ -20,14 +20,16 @@ import {
 	handleCrossReference
 } from "../internal"
 
-declare global {
-	// `Iterator.from` was added in ES2025.
-	var Iterator:
-		| undefined
-		| {
-				from<T, TReturn>(iterable: Iterator<T, TReturn>): IterableIterator<T>
-		  }
-}
+// Feature-detect `Iterator.from` (added in ES2025) without augmenting the
+// global scope. Declaring `var Iterator` in a `declare global` block leaks a
+// conflicting global variable into the bundled `.d.ts` (TS2403 when compiled
+// against the ESNext lib). We cast via `globalThis` to satisfy the type checker
+// while keeping the runtime feature detect accurate. (#1273)
+const _globalIterator = (globalThis as any).Iterator as
+	| undefined
+	| { from: <T>(iterable: Iterator<T>) => IterableIterator<T> }
+const hasIteratorFrom =
+	typeof _globalIterator?.from === "function"
 
 export function enableMapSet() {
 	class DraftMap extends Map {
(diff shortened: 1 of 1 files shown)


mweststrate:
Looks great, thanks for addressing!

Issue #1143, opened 2024-09-09
CI Build issue patch


nishant8BITS:
Hi! 👋 

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/immer/src/core/immerClass.ts b/node_modules/immer/src/core/immerClass.ts
index f827361..37bb8c9 100644
--- a/node_modules/immer/src/core/immerClass.ts
+++ b/node_modules/immer/src/core/immerClass.ts
@@ -1,30 +1,30 @@
 import {
-	IProduceWithPatches,
+	DRAFT_STATE,
+	Draft,
+	Drafted,
 	IProduce,
+	IProduceWithPatches,
 	ImmerState,
-	Drafted,
-	isDraftable,
-	processResult,
-	Patch,
+	NOTHING,
 	Objectish,
-	DRAFT_STATE,
-	Draft,
+	Patch,
 	PatchListener,
-	isDraft,
-	isMap,
-	isSet,
 	createProxyProxy,
-	getPlugin,
+	current,
 	die,
 	enterScope,
-	revokeScope,
-	leaveScope,
-	usePatchesInScope,
-	getCurrentScope,
-	NOTHING,
 	freeze,
-	current
-} from "../internal"
+	getCurrentScope,
+	getPlugin,
+	isDraft,
+	isDraftable,
+	isMap,
+	isSet,
+	leaveScope,
+	processResult,
+	revokeScope,
+	usePatchesInScope
+} from "../internal";
 
 interface ProducersFns {
 	produce: IProduce
@@ -178,7 +178,7 @@ export class Immer implements ProducersFns {
 		let i: number
 		for (i = patches.length - 1; i >= 0; i--) {
 			const patch = patches[i]
-			if (patch.path.length === 0 && patch.op === "replace") {
+			if (patch?.path?.length === 0 && patch?.op === "replace") {
 				base = patch.value
 				break
 			}
diff --git a/node_modules/immer/src/utils/common.ts b/node_modules/immer/src/utils/common.ts
index 692157b..8db10ba 100644
--- a/node_modules/immer/src/utils/common.ts
+++ b/node_modules/immer/src/utils/common.ts
@@ -1,14 +1,14 @@
 import {
-	DRAFT_STATE,
-	DRAFTABLE,
-	Objectish,
-	Drafted,
-	AnyObject,
 	AnyMap,
+	AnyObject,
 	AnySet,
-	ImmerState,
 	ArchType,
 	die,
+	DRAFT_STATE,
[...]

PR #1297, opened 2026-09-05
fix: guard indexed access in applyPatches and shallowCopy under noUncheckedIndexedAccess

nirmal-shaji:
## Problem

Enabling TypeScript's [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) breaks the build with `TS18048: '...' is possibly 'undefined'` in two spots that read a value straight out of an indexed access:

- `Immer.applyPatches` (`src/core/immerClass.ts`) — `const patch = patches[i]`, then `patch.path` / `patch.op` / `patch.value`.
- `shallowCopy` (`src/utils/common.ts`) — `const desc = descriptors[key]`, then `desc[WRITABLE]` / `desc.get` / `desc.set` / `desc[ENUMERABLE]`.

Reported in #1143.

## Cause

Under `noUncheckedIndexedAccess`, `arr[i]` / `record[key]` are typed as `T | undefined`, so accessing members of the result without first narrowing it is a compile error. The indices are in fact always in range here (a reverse loop bounded by `patches.length`, and keys taken from `Reflect.ownKeys(descriptors)`), but the compiler cannot know that.

## Fix

Guards only — **no runtime behavior change.** Add an early `if (!patch) continue` / `if (!desc) continue` immediately after each indexed read. This narrows the value to a non-undefined type for the rest of the loop body, satisfying the compiler. Because the indices are always valid at runtime, the guard condition can never be true and the `continue` never executes, so existing behavior is preserved exactly.

```diff
 		const patch = patches[i]
+		if (!patch) continue
 		if (patch.path.length === 0 && patch.op === "replace") {
```

```diff
 		const desc = descriptors[key]
+		if (!desc) continue
 		if (desc[WRITABLE] === false) {
```

An early guard is used rather than per-access optional chaining because it also covers the later `patch.value` read / the descriptor rewrite in one place, and reads more clearly.

## Testing

- `yarn test:src` (vitest) — **3764 passed, 8 skipped**, full suite green.
- Type check: compiled the source entry points with `noUncheckedIndexedAccess` enabled (a tsconfig extending the repo's own `tsconfig.json`). **Before:** 9 `TS18048` errors across these two files; **after:** 0 — both files compile clean under strict indexed access. The default `tsconfig.json` build is unaffected.

> Note: this PR is scoped to the two locations #1143 identifies. `src/plugins/patches.ts` has a few remaining, different indexed-access errors (`TS2345`/`TS2538`) that would need to be addressed before the whole codebase compiles under `noUncheckedIndexedAccess`; happy to follow up separately if the maintainers want full opt-in.


markerikson:
That sounds like your TS is trying to check code _inside_ Immer, and checking library code while building your app is generally a bad idea.  Is there a reason why you're doing that?

nirmal-shaji:
Fair point. It only bites when Immer's own `src` is compiled under `noUncheckedIndexedAccess`, which isn't a normal consumer setup. It's guards only, no runtime change, so there's no strong case to keep it. Your call.

nirmal-shaji:
Closing this, not worth the maintenance if the source isn't built under that flag. Thanks for taking a look.

Issue #1087, opened 2024-01-06
`produce` leaks proxy objects when symbols are present


dhmk083:
## 🐛 Bug Report

When you use `produce` on an object which contains a symbol property which is an object, it leaks `Proxy` object into returned value.

## Link to repro

[https://codesandbox.io/p/sandbox/suspicious-sanne-2ypqwt](https://codesandbox.io/p/sandbox/suspicious-sanne-2ypqwt?file=%2Fsrc%2Findex.ts)

## To Reproduce

1. Get an object with symbol:

```js
const sym = Symbol()

let state = {
  id: 1,
  [sym]: {} // must be an object
}
```

2. Use it with `produce` and read `state[sym]` value inside draft function. In returned value `state[sym]` object will be a revoked proxy instead of plain object.

3. Use it with `produce` again. Now an error will be thrown because `state[sym]` is corrupted.

```
TypeError
Cannot perform 'getPrototypeOf' on a proxy that has been revoked
```

## Observed behavior

Proxy object is leaking out of `produce` function.

## Expected behavior

Plain object should be returned instead.

## Environment

The bug appears since 10.0.0 version. It works fine in 9.0.21 version.

mweststrate:
Is it fixed when enabling useStrictShallowCopy(true)?

dhmk083:
No, still error.

lsby:
I encountered the same issue. Reverting to version 9.0.21 resolved the problem for me as well.

joshkel:
This was changed here: https://github.com/immerjs/immer/commit/452a5085c0f84c3277ee16ab9b072ad99127ce90#diff-4ade25325a6cd7d1cca619364c84d19bfa47e6939a2c1b38d065767255b8f86eR66

Based on the commit messages, it looks like it might have been a mistake (a change made to disable a non-default flag, but the behavior was indadvertently changed), but I could be wrong.

lveillard:
Hello! Had the same issue, objects nested on symbols are automatically revoked.

What older version of immer could we use meanwhile?

mweststrate:
Likely same issue as #1096. Will investigate when I can carve out a sliver of time. If someone is willing to submit a PR with  unit test(s), that will already be quite helpful!

mweststrate:
#1105 should fix the handling of Proxy properties. When drafted, they are now finalized correctly. Let me know if it isn't solved by `immer@10.0.4`

PR #1306, opened 2026-09-20
fix: stack overflow on pre-existing Symbol circular references (#1106)

### LINKED ISSUE #1106: "Maximum call stack exceeded" with circular references
## 🐛 Bug Report

Starting with Immer 10.0.4, if Immer is given an object that uses symbols for circular references, it generates a "Maximum call stack exceeded" error.

## Link to repro

https://playcode.io/1794809

## To Reproduce

Pass an object with pre-existing symbol circular references to Immer 10.0.4. (Circular references that are added within an Immer recipe do not cause problems.)

## Observed behavior

"Maximum call stack exceeded"

## Expected behavior

Our application uses symbols to create child-to-parent relationships. Since symbols are skipped by many JavaScript routines (e.g., JSON.stringify, Immer prior to 10.0.4), we had been able to do this without error.

## Environment

- **Immer version:** 10.0.4
- [X] I filed this report against the _latest_ version of Immer

dyk1454683243-sudo:
Fixes #1106.

## Problem

Starting in 10.0.4, Immer walks symbol keys. Objects that use symbols for child-to-parent links (skipped by `JSON.stringify` / Immer 9) then hit **Maximum call stack size exceeded**.

On current `main`, `produce()` itself is already safe: `handleValue` tracks visited objects. `current()` was not. It:

1. Recursed into nested non-drafts with **no cycle guard**.
2. Always used **strict** iteration for those nodes (`strict` was not passed down), so symbol links were followed even when `setUseStrictIteration(false)` (the default).

That is the remaining overflow for the issue’s data shape — e.g. `produce(parent, d => { d.name = "x"; current(d) })`.

## Change

In `current()`:

- Remember objects already snapshotted (`WeakMap`) so circular refs (including symbol keys) reuse the snapshot instead of recursing forever.
- Pass the same iteration mode into nested non-drafts.

No other finalization / copy behavior is changed.

## Tests

`__tests__/symbol-circular.js` covers the issue’s cases:

- `produce` on parent, nested child, and child-as-root
- reading a pre-existing symbol without leaking drafts
- `current()` after parent/child updates
- self-referential symbol props
- `setUseStrictIteration(true)` and strict shallow copy
- symbols assigned *inside* a recipe (still no leaked drafts; #1087)

## Notes

Circular references added *inside* a recipe were already fine; this is only about **pre-existing** symbol cycles. Plain (enumerable) circular refs remain unsupported, as documented.


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

